PR #3, opened 2023-11-07
Arm Cartesian Work

Coolearthsky:


studies/armCartesianSpaceSquare/src/main/java/frc/robot/RobotContainer.java
```diff
@@ -22,7 +22,12 @@ private void configureBindings() {
     public Command getAutonomousCommand() {
         return m_auton;
     }
-
+    public void runPeriodic() {
```
truher, line 25: i'll merge this as is, but you might want to move these logging statements to the subsystem periodic method instead of creating a periodic in RobotContainer.

studies/armCartesianSpaceSquare/src/main/java/frc/robot/arm/ArmTrajectory.java
truher, line 1: somewhere in every command you should "require" the subsystem you're using, but i can't find where you do that.

truher (approved):

PR #4, opened 2023-11-07
initial commit. includes the backslash in arm sub

truher:


lib/src/main/java/org/team100/lib/motion/drivetrain/SwerveLocal.java
```diff
@@ -61,18 +62,20 @@ public SwerveLocal(
      * @param targetChassisSpeeds speeds in robot coordinates.
      */
     public void setChassisSpeeds(ChassisSpeeds targetChassisSpeeds) {
-        if (m_experiments.enabled(Experiment.UseSetpointGenerator)) {
+        // if (m_experiments.enabled(Experiment.UseSetpointGenerator)) {
```
truher, line 65: i've noticed you have a habit of commenting out the experiment conditionals.  i'll merge this as is but maybe next time just use the experiment mechanism as designed?  like in this case i think the setpoint generator experiment is enabled, so it should be running already.

PR #20, opened 2023-11-11
Subsystem arm

Coolearthsky:


studies/armCartesianSpaceSquare/src/main/java/frc/robot/subSystems/FRCNEO.java
```diff
@@ -1,4 +1,4 @@
-package frc.robot;
+package frc.robot.subSystems;
```
truher, line 1: package names should be all lower case.  also if you wanted to follow the lib example this would be in package named something like org.team100.cartesiansquare.motion.arm

lib/src/main/java/org/team100/lib/hid/JoystickControl.java
```diff
@@ -49,7 +49,7 @@ public Twist2d twist() {
                 deadband(-1.0 * clamp(m_controller.getY(), 1), m_config.kDeadband, 1),
                 m_config.kExpo);
         double dy = expo(
-                deadband(-1.0 * clamp(m_controller.getX(), 1), m_config.kDeadband, 1),
+                deadband(1.0 * clamp(m_controller.getX(), 1), m_config.kDeadband, 1),
```
truher, line 52: so if FRC positive Y is to the left, and Xbox positive Y is to the right, so i think the inversion you removed was actually correct.  in the past we've made this mistake (several times) because we do things like map the corners of the drivetrain incorrectly or something, and inverting the control "fixes" it.  but it's not fixed.

lib/src/main/java/org/team100/lib/hid/DualXboxControl.java
```diff
@@ -86,7 +86,7 @@ public void resetRotation180(Command command) {
     @Override
     public Twist2d twist() {
         double dx = expo(deadband(-1.0 * clamp(controller0.getRightY(), 1), m_config.kDeadband, 1), m_config.kExpo);
-        double dy = expo(deadband(-1.0 * clamp(controller0.getRightX(), 1), m_config.kDeadband, 1), m_config.kExpo);
+        double dy = expo(deadband(1.0 * clamp(controller0.getRightX(), 1), m_config.kDeadband, 1), m_config.kExpo);
```
truher, line 89: see my other comment; i think this is wrong

studies/armCartesianSpaceSquare/src/main/java/frc/robot/armCommands/ArmTrajectoryCommand.java
```diff
@@ -0,0 +1,58 @@
+package frc.robot.armCommands;
```
truher, line 1: use lower case package names

studies/armCartesianSpaceSquare/src/main/java/frc/robot/armCommands/ArmTrajectoryCommand.java
```diff
@@ -0,0 +1,58 @@
...
+
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.Command;
+import frc.robot.subSystems.ArmSubsystem;
+
+public class ArmTrajectoryCommand extends Command {
+
+    private boolean t = false;
```
truher, line 10: use a more descriptive variable name

studies/armCartesianSpaceSquare/src/test/java/frc/test2.java
```diff
@@ -5,7 +5,7 @@
 import org.junit.jupiter.api.Test;
 
 import edu.wpi.first.math.trajectory.TrajectoryConfig;
-import frc.robot.arm.ArmTrajectories;
+import frc.robot.armMotion.ArmTrajectories;
 
 public class test2 {
```
truher, line 10: maybe name this something more descriptive?

truher (changes requested):

Coolearthsky:
Fixed

truher (approved):

Issue #24, opened 2023-11-14
use measured velocity for stiction control, not setpoint velocity


truher:
https://github.com/Team100/all24/blob/d41b80609522b0eb3eb8ef550acfd0d0e55ae9a7/lib/src/main/java/org/team100/lib/motor/drive/FalconDriveMotor.java#L108C23-L108C23

PR #25, opened 2023-11-15
Bug fix to drive module, IMPORTANT: tune module

Coolearthsky:


truher:
i believe this fixes #24

PR #30, opened 2023-11-18
Code fix to swerve 2

Coolearthsky:


812405:
try to make a little more specific in the future i.e. "reversed wheel direction"

PR #35, opened 2023-11-19
Object Detection

Coolearthsky:


truher (changes requested):
looks like all the tensorflow parts are commented out?  did you mean to do that?

Coolearthsky:
I got rid of all the old barcode testing stuff with averages that turned out to not be useful.

truher:
> I got rid of all the old barcode testing stuff with averages that turned out to not be useful.

ok, thanks.  since i made the changes in a PR 50 your changes here will end up as a merge conflict. you can remove them, they'll go in with PR 50.

there's one more question about the thresholds; how did you determine what they should be?  how sensitive is the threshold?  one is 2.4 and another is 2.5 which implies that the level of 0.1 is important?

truher:
could you add a comment somewhere in the barcode code that describes what
the useful changes actually were?  you worked on it for a long time, i'd
like there to be record of what you learned.

PR #50, opened 2023-11-23
barcode to comp

truher:


lib/src/main/java/org/team100/lib/barcode/Sensor.java
```diff
@@ -0,0 +1,256 @@
...
+ * 
+ * 
+ * Pololu product page:
+ * 
+ * https://www.pololu.com/product/4405/resources
+ */
+public class Sensor {
+    private static final int loops = 150;
+
+    private final double[] s5 = new double[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
truher, line 67: i think you can replace this with new double[150], unless you want to specify something other than all zeros?

lib/src/main/java/org/team100/lib/barcode/Sensor.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    /**
+     * @param threshold above the threshold means dark, 1, below means light, is 0
+     * @return true = dark = "1" in the bar code.
+     */
+    boolean readInput(double threshold, int i) {
+        double sv = m_input.getAverageVoltage();
+        switch (i) {
+            case 0:
```
truher, line 187: could you explain a little about what you're doing here?  it looks like you're averaging over a sliding window with 150 samples?  could you use the WPILib filter for that?

lib/src/main/java/org/team100/lib/barcode/Barcode.java
```diff
@@ -0,0 +1,85 @@
...
+                new DigitalOutput(2),
+                new DigitalOutput(3)
+        };
+        mux = new Mux(m_outputs);
+        m_input = new AnalogInput(0);
+        m_odd = new DigitalOutput(5);
+        m_even = new DigitalOutput(4);
+
+        // TODO: tune the thresholds
+        double[] thresholds = new double[] { 2.5, 3, 3, 2.4, 2.5 };
```
truher, line 33: is there something behind these numbers? is there some range of numbers that work, and other numbers that don't work?

truher (commented):
hey, before i merge the barcode work i have a few questions about it.  it's a little bit awkward because this PR contains both the "study" version and the "comp" version and it's in my fork so you can't actually change anything, so maybe just reply inline and i'll make whatever changes you want?

PR #118, opened 2024-01-17
PositionServo Indexer

Coolearthsky:


comp/swerve100/src/main/java/org/team100/frc2024/motion/indexer/IndexerSubsystem.java
```diff
@@ -1,26 +1,64 @@
...
     public void set(double value) {
         setpoint = m_trapezoid.calculate(0.02, setpoint, new State100(value, 0, 0));
-        m_motor.setVelocity(setpoint.x(), 0);
+        m_servo.setVelocity(setpoint.x());
+    }
+
+    public void setPosition(double value) {
+        m_servo.setPosition(value);
+    }
+    private static PositionServo<Angle> newPositionServo(
```
truher, line 38: i think you want PositionServo<Distance> here since the indexer is actually providing linear actuation of the game piece.

see ServoFactory.

truher (changes requested):

PR #125, opened 2024-01-18
Changes to subsystems and to have the code work with different subsystems (alpha and beta bot). Position Servos were added to all the subsystems and the trapezoid profile was moved from the subsystems to the position servo

anorakthegreat:


truher (changes requested):
this is a good start!   i think you're using PositionServo incorrectly, though: it's supposed to manage the position of something, and the fact that it has a velocity passthrough is really just for testing.

I think what you're aiming for is a class that sits between the command and the outboard velocity controller: command sets goals, new thing smooths them, velocityservo executes the smoothed setpoints.  We don't really have anything like that, so maybe make one?  Just wrap the velocityservo and use a profile. 

i like the configurable shooter factory, good idea.  using the Identity to select would be even better; the pattern is "switch(Identity.instance)" and you can see it in SwerveModuleCollection among other places.

PR #148, opened 2024-01-22
Merge code changes into main

Coolearthsky:


comp/swerve100/src/main/java/org/team100/frc2024/motion/indexer/IndexCommand.java
truher, line 1: i like the idea of this command.  i think it would be cleaner if the indexer didn't know about the shooter velocity at all, or even what it was waiting for -- it could just take a boolean supplier.  then you could have something else (e.g. the shooter) supply "true" when it was ready to shoot.

comp/swerve100/src/main/java/org/team100/frc2024/RobotContainer.java
```diff
@@ -317,14 +318,13 @@ public RobotContainer(TimedRobot robot) throws IOException {
         // TODO: intake whenever intake is running.
         // TODO: stop when note is accpeted using optical detector.
         // TODO: shoot only when the shooter is ready.
-
         m_indexer.setDefaultCommand(m_indexer.run(m_indexer::stop));
-        operatorControl.index().whileTrue(m_indexer.run(m_indexer::forward));
+        //TODO get real desired velocity before indexing
+        operatorControl.index().whileTrue(new IndexCommand(m_amp, m_indexer, m_shooter, 29));
```
truher, line 323: when you're working on a new command, don't replace the old behavior until you're completely finished.  make a new operator control trigger, and keep your work separate from everything else until it's done.

truher (changes requested):
lots of merge conflicts here

PR #160, opened 2024-01-27
Outboard kV tunable

Coolearthsky:


lib/src/main/java/org/team100/lib/motor/drive/NeoDriveMotor.java
```diff
@@ -76,7 +76,9 @@ public NeoDriveMotor(
             boolean motorPhase,
             int currentLimit,
             double gearRatio,
-            double wheelDiameter) {
+            double wheelDiameter,
+            double kV) {
+        outboardP = kV;
```
truher, line 81: P and kV are not the same idea

lib/src/main/java/org/team100/lib/motor/turning/NeoTurningMotor.java
```diff
@@ -72,7 +72,8 @@ public class NeoTurningMotor implements Motor100<Angle100> {
     /** Current position measurement, obtained in periodic(). */
     private double m_encoderPosition;
 
-    public NeoTurningMotor(String name, int canId, boolean motorPhase, int currentLimit, double gearRatio) {
+    public NeoTurningMotor(String name, int canId, boolean motorPhase, int currentLimit, double gearRatio, double kV) {
+        outboardP = kV;
```
truher, line 76: P and kV are not the same idea

truher (changes requested):

PR #163, opened 2024-01-28
rear right wheel needs to be looked at, PID controller widgets added

anorakthegreat:


lib/src/main/java/org/team100/lib/motion/drivetrain/module/WCPSwerveModule100.java
```diff
@@ -108,20 +110,21 @@ private static PositionServoInterface<Angle100> turningServo(
...
             double gearRatio,
-            SwerveKinodynamics kinodynamics) {
+            SwerveKinodynamics kinodynamics,
+            Drive drive) {
         final double turningGearRatio = 1.0;
         FalconTurningMotor turningMotor = new FalconTurningMotor(
                 name,
                 turningMotorCanId,
-                true,
+                false,
```
truher, line 119: this should be configurable; if you just say "false" here then it breaks the other robot

lib/src/main/java/org/team100/lib/motor/drive/NeoDriveMotor.java
```diff
@@ -77,7 +79,8 @@ public NeoDriveMotor(
             int currentLimit,
             double gearRatio,
             double wheelDiameter,
-            double kV) {
+            double kV,
+            PIDController lowLevelVelocityController) {
```
truher, line 83: don't pass a *controller* here since you're not using it to control anything, and you're also not listening for updates to the parameters through smart dashboard.  just pass the coefficients, or make a config object for them like your SysParam idea.

lib/src/main/java/org/team100/lib/motor/turning/NeoTurningMotor.java
```diff
@@ -72,7 +74,7 @@ public class NeoTurningMotor implements Motor100<Angle100> {
     /** Current position measurement, obtained in periodic(). */
     private double m_encoderPosition;
 
-    public NeoTurningMotor(String name, int canId, boolean motorPhase, int currentLimit, double gearRatio, double kV) {
+    public NeoTurningMotor(String name, int canId, boolean motorPhase, int currentLimit, double gearRatio, double kV, PIDController lowLevelVelocityController) {
```
truher, line 77: same story with this controller.  use a config object not a controller object.

truher (changes requested):

PR #164, opened 2024-01-28
rear right needs to be looked at, added MotorPhase enum to turning motors

anorakthegreat:


comp/swerve100/src/main/java/org/team100/frc2024/motion/amp/AmpSubsystem.java
```diff
@@ -76,6 +96,7 @@ public AmpSubsystem(int leftPivotID, int rightPivotID) {
      * @param value
      */
     public void setAmpPosition(double value) {
+        System.out.println("AHHHHHHHHHHHH");
```
truher, line 99: remove

Coolearthsky, line 99: Could you make this a log with a nicer name?

lib/src/main/java/org/team100/lib/config/PIDConstants.java
```diff
@@ -0,0 +1,53 @@
+package org.team100.lib.config;
+
+import edu.wpi.first.wpilibj.DigitalInput;
+import edu.wpi.first.wpilibj.DriverStation;
+
+/** Uses digital inputs 4 and 5. */
```
truher, line 6: this comment doesn't make sense to me.

also you could make this a "record" and it would be cleaner

truher (approved):

PR #171, opened 2024-01-30
Gamepiece detector

Coolearthsky:


comp/vision/gamepiece_finder24.py
```diff
@@ -50,7 +60,7 @@ def initialize_nt(self):
 
         # work around https://github.com/robotpy/mostrobotpy/issues/60
         self.inst.getStructTopic("bugfix", NotePosition).publish().set(
-            NotePosition(Translation2d())
+            NotePosition(int(),int())
```
truher, line 63: instead of "int()" here i think you just mean "0"

comp/vision/gamepiece_finder24.py
```diff
@@ -73,13 +83,17 @@ def find_object(self, img):
         for c in contours:
             _, _, cnt_width, cnt_height = cv2.boundingRect(c)
             if (cnt_height < 50):
+                print("WWWW")
```
truher, line 86: maybe take out all these print statements

lib/src/main/java/org/team100/lib/config/FeedforwardConstants.java
```diff
@@ -6,9 +6,9 @@ public class FeedforwardConstants {
     private final double m_kSS;
     private final double m_kDS;
 
-    //Default no friction or attached objects contraints
+    //Default no friction or attached objects contraints for NEO
     public FeedforwardConstants() {
```
truher, line 10: don't make this a constructor, make a static method that says what it means, e.g. "public static FeedforwardConstants makeNeo() {return new FeedfowardConstants(0.122,0,0.1,0.065);}"

truher (changes requested):
looks good, i made a few suggestions.  one other suggestion is to get rid of the cropping on line 133 of gamepiece_finder24.py; you want the maximum vertical size, i think.

lib/src/main/java/org/team100/lib/config/FeedforwardConstants.java
Coolearthsky: I don't know why this still says it's not outdated but I fixed this

PR #178, opened 2024-02-05
Object detection stuff

Coolearthsky:
Don't merge it yet, im going to add a profile to the note drive command


comp/swerve100/src/main/java/org/team100/frc2024/RobotContainer.java
```diff
@@ -267,6 +275,10 @@ public RobotContainer(TimedRobot robot) throws IOException {
         whileTrue(driverControl::never,
                 new DriveToWaypoint100(goal, m_drive, planner, drivePID, swerveKinodynamics));
 
+        //Drive With Profile
+        whileTrue(operatorControl::driveToNote,
+                new DriveWithProfile(() -> m_noteCamera.fieldRelativePose2d(), m_drive, dthetaController, swerveKinodynamics));
```
truher, line 280: instead of a trivial lambda you can just refer to the function:

m_noteCamera::fieldRelativePose2d

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile.java
```diff
@@ -0,0 +1,105 @@
...
+        State100 thetaGoal = new State100(bearing.getRadians(), 0);
+        State100 xGoalRaw = new State100(m_robotRelativeGoal.get().getX(),0,0);
+        xSetpoint = xProfile.calculate(0.02, xSetpoint, xGoalRaw);
+        State100 yGoalRaw = new State100(m_robotRelativeGoal.get().getY(),0,0);
+        ySetpoint = yProfile.calculate(0.02, ySetpoint, yGoalRaw);
+        // State100 thetaGoalRaw = new State100(m_robotRelativeGoal.get().getRotation().getRadians(),0,0);
+        thetaSetpoint = thetaProfile.calculate(0.02, thetaSetpoint, thetaGoal);
+        SwerveState goalState = new SwerveState(xSetpoint, ySetpoint, thetaSetpoint);
+        Twist2d goal = m_controller.calculate(m_swerve.getState(), goalState);
+        m_swerve.driveInFieldCoords(goal, 0.02);
```
truher, line 88: this function seems to be taking a "robot-relative goal" and feeding it into the controller which is expecting a field-relative goal.

lib/src/main/java/org/team100/lib/hid/OperatorControl.java
```diff
@@ -26,6 +26,10 @@ default boolean pivotToAmpPosition(){
         return false;
     }
 
+    default boolean driveToNote() {
```
truher, line 29: why do you want both the driver and the operator to be able to trigger the drive-to-note behavior?

lib/src/main/java/org/team100/lib/localization/Blip24ArrayListener.java
```diff
@@ -42,7 +42,7 @@ void consumeValues(NetworkTableEvent e) {
                 System.out.println(fields[1] + " " + blip);
             }
         } else {
-            System.out.println("weird vision update key: " + name);
+            // System.out.println("weird vision update key: " + name);
```
truher, line 45: you should probably look at why this is happening, instead of just commenting it out.

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -32,24 +38,53 @@ void consumeValues(NetworkTableEvent e) {
...
+     *         screen
+     */
+    public Optional<Double> getX() {
+        switch (Identity.instance) {
+            case BETA_BOT:
+                double xd = positions[0].getX();
+                Double x = xd;
+                Optional<Double> e = Optional.of(x);
+                return e;
+            default:
```
truher, line 67: i think what you're supposed to do here is use Optional.empty() as the default value.

lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -0,0 +1,197 @@
...
+     *         translation and radians for rotation
+     */
+    public Pose2d fieldRelativePose2d() {
+        switch (Identity.instance) {
+            case BETA_BOT:
+                if (m_notePosition24ArrayListener.getY() != null && m_notePosition24ArrayListener.getX() != null) {
+                    return m_swerve.getPose().transformBy(robotRelativeTransform2d());
+                }
+                return null;
+            default:
```
truher, line 190: the default should never seem like real data.

lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -0,0 +1,197 @@
...
+import org.team100.lib.motion.drivetrain.SwerveDriveSubsystem;
+import org.team100.lib.telemetry.Telemetry;
+import org.team100.lib.telemetry.Telemetry.Level;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.geometry.Transform2d;
+import edu.wpi.first.math.geometry.Translation2d;
+
+public class CameraAngles {
```
truher, line 14: it would be good for this class to *just* do the math part, and not know about things like the listener or the swerve drive.  put that stuff somewhere else.

lib/src/test/java/org/team100/lib/util/PerspectiveTest.java
```diff
@@ -38,13 +38,11 @@ Mat calibrationSquareInCameraPixels() {
...
         // width is 832 so center is 416
         // height is 616 so center is 308.
-        // these are guesses, completely different from the actual camera
-        // TODO: do a real calibration; note it depends on the camera angle.
         List<Point> srcpts = new ArrayList<Point>();
-        srcpts.add(new Point(366, 666));
-        srcpts.add(new Point(466, 666));
-        srcpts.add(new Point(316, 566));
-        srcpts.add(new Point(516, 566));
+        srcpts.add(new Point(389, 643));
```
truher, line 42: add a comment describing where these numbers came from

comp/swerve100/src/main/java/org/team100/frc2024/RobotContainer.java
```diff
@@ -187,6 +191,7 @@ public RobotContainer(TimedRobot robot) throws IOException {
                 poseEstimator,
                 swerveLocal,
                 driverControl::speed);
+        m_noteCamera = new CameraAngles(30, 67.5, 50, 832, 616, 0.71, notePositionDetector, m_drive);
```
truher, line 194: are you using the v2 camera or the v3 camera?  they have different FOV.  you could configure the camera type in the Camera class (which is like Identity for cameras) or you could pass the camera type through network tables, it's available in python as camera.camera_properties['Model'].

truher (changes requested):

PR #181, opened 2024-02-06
PP works with choreo, shooter lock added

anorakthegreat:


truher (changes requested):
looks like the build of this PR failed?  is that a real failure?

anorakthegreat:
It builds on my computer fine?? No clue why its mad

truher:
i'll merge it and see

PR #188, opened 2024-02-07
Fixed note facing commands

Coolearthsky:


lib/src/main/java/org/team100/lib/motion/drivetrain/manual/ManualWithNoteRotation.java
```diff
@@ -92,8 +94,9 @@ public void reset(Pose2d currentPose) {
      * 
      * @param state from the drivetrain
      * @param input control units [-1,1]
-     * @return feasible field-relative velocity in m/s and rad/s
+     * @return Working on this, theta overshoots for some reason
```
truher, line 97: why did you remove the comment about the return value?

lib/src/main/java/org/team100/lib/config/NoteDetector.java
```diff
@@ -104,7 +104,8 @@ public Pose2d fieldRelativePose2d() {
             case BETA_BOT:
                     return m_swerve.getPose().transformBy(robotRelativeTransform2d());
             default:
-                return new Pose2d();
+            //I want the sim to go to a pose similar to how it would attack a note, so I am putting the angle of attack equal to 0,0 with the angle to that point, is there a better way to do this?
```
truher, line 107: for simulation don't you want to have, like, a simulated note position?

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveManually.java
```diff
@@ -123,6 +123,21 @@ public void reset(Pose2d p) {
                 });
     }
 
+    public void register(String name, boolean isDefault, WeirdChassisSpeedDriver d) {
```
truher, line 126: remove this, extend ChassisSpeedDriver instead.

comp/swerve100/src/main/java/org/team100/frc2024/motion/drivetrain/ShooterUtil.java
```diff
@@ -7,25 +7,23 @@
...
 
-import edu.wpi.first.math.MathUtil;
 import edu.wpi.first.math.geometry.Translation2d;
 
-/** Add your docs here. */
 public class ShooterUtil {
 
-    public static Translation2d getOffsetTranslation(SwerveState state, double kScale){
-        Translation2d currentTranslation = state.pose().getTranslation();
+    public static Translation2d getOffsetTranslation(SwerveState state, double shooterVelocity){
```
truher, line 14: why is this change in here?

truher (changes requested):

comp/swerve100/src/main/java/org/team100/frc2024/motion/drivetrain/ShooterUtil.java
anorakthegreat: Make a new method for this. One attempts to address shooting in the corner of the shooter and the other attempts to address shooting while moving

PR #191, opened 2024-02-07
unfinished shooting while moving and amp trajec code

anorakthegreat:


truher:
merged.  i think if you sync you should see these changes.  it's weird that the github gui doesn't think it's done but it looks done to me.

PR #197, opened 2024-02-08
Falcon6 merge and more

Coolearthsky:


lib/src/main/java/org/team100/lib/config/NoteDetector.java
```diff
@@ -35,23 +36,42 @@ public NoteDetector(
...
     public double getAngleToNote() {
-        return m_cameraAngles.getAngle(m_notePosition24ArrayListener.getX().get(),
-                m_notePosition24ArrayListener.getY().get());
+        switch (Identity.instance) {
+            case BETA_BOT:
+            case COMP_BOT:
+                return m_cameraAngles.getAngle(m_notePosition24ArrayListener.getX().get(),
+                        m_notePosition24ArrayListener.getY().get());
+            case BLANK:
+                return MathUtil.angleModulus(m_swerve.getPose().getTranslation().getAngle().getRadians() + Math.PI);
```
truher, line 70: what is this expression supposed to mean?  it seems like the angle to the origin (i.e. corner of the field), plus 90 degrees?

lib/src/main/java/org/team100/lib/config/NoteDetector.java
```diff
@@ -68,43 +88,46 @@ public Double getY() {
...
      *         meters
      */
     public Translation2d FieldRelativeTranslation2d() {
-            return new Translation2d(fieldRelativeX(), fieldRelativeY());
+        return new Translation2d(fieldRelativeX(), fieldRelativeY());
     }
 
     /**
      * @return A robot relative Transform2d value of an object in a camera in meters
      *         for translation and radians for rotation
```
truher, line 120: maybe add a bit to this comment pointing out the rotational value of this pose, which is the bearing.

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -231,12 +228,13 @@ public void setVelocity(double outputM_S, double accelM_S_S) {
...
         v.FeedForward = kFF;
-        v.Acceleration = motorRev_S2;
         v.EnableFOC = true;
+        v.withFeedForward(kFF);
+        v.Acceleration = motorRev_S2;
         m_motor.setControl(v);
 
         // m_motor.set(ControlMode.Velocity, motorTick_100ms, DemandType.ArbitraryFeedForward, kFF);
-
+        t.log(Level.DEBUG, m_name, "motor input [-1,1]", motorRev_S);
```
truher, line 237: i think you want rev_s in the name of this item, not -1,1

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -231,12 +228,13 @@ public void setVelocity(double outputM_S, double accelM_S_S) {
 
         VelocityDutyCycle v = new VelocityDutyCycle(motorRev_S);
         v.FeedForward = kFF;
-        v.Acceleration = motorRev_S2;
         v.EnableFOC = true;
+        v.withFeedForward(kFF);
```
truher, line 232: why are you using this method instead of setting the field?

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -282,7 +280,7 @@ public void resetPosition() {
     public void periodic() {
         // m_rawPosition = m_motor.getSelectedSensorPosition();
         m_rawPosition = m_motor.getPosition().getValueAsDouble();
-        m_velocityRev_S = m_motor.getVelocity().getValueAsDouble();
+        m_velocityRev_S = m_motor.getVelocity().getValueAsDouble()/m_gearRatio;
```
truher, line 283: it seems like you're redefining the meaning of the m_velocityRev_S field.  it used to be velocity of the motor; now you're saying it's velocity of the output shaft, but this meaning is inconsistent with everything else in this class.

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -31,39 +31,37 @@ public class Falcon6DriveMotor implements MotorWithEncoder100<Distance100> {
      * The speed, below which, static friction applies, in motor revolutions per
      * second.
      */
-    private static final double staticFrictionSpeedLimitRev_S = 3.5;
+    private static final double staticFrictionSpeedLimitRev_S = 0.1;
```
truher, line 34: i'm pretty sure the large value that used to be here was tuned on the ground, and i think you tuned this on blocks.  also this is affected by the other bug about gear ratio, see below.

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -31,39 +31,37 @@ public class Falcon6DriveMotor implements MotorWithEncoder100<Distance100> {
...
-    private static final double dynamicFrictionFFVolts = 0.01;
+    private static final double dynamicFrictionFFVolts = 0.019;
 
     /**
      * Velocity feedforward in units of volts per motor revolution per second, or
      * volt-seconds per revolution. Since saturation is 11 volts and free speed is
      * about 100 rev/s, this is about 0.11.
      */
-    private static final double velocityFFVoltS_Rev = 0.11;
+    private static final double velocityFFVoltS_Rev = 0.06;
```
truher, line 55: this value seems wrong, and it's affected by the volt/duty-cycle bug below.

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -31,39 +31,37 @@ public class Falcon6DriveMotor implements MotorWithEncoder100<Distance100> {
...
      */
-    private static final double staticFrictionSpeedLimitRev_S = 3.5;
+    private static final double staticFrictionSpeedLimitRev_S = 0.1;
 
     /**
      * Friction feedforward in volts, for when the mechanism is stopped, or nearly
      * so.
      */
-    private static final double staticFrictionFFVolts = 0.18;
+    private static final double staticFrictionFFVolts = 0.03;
```
truher, line 40: same issue with static friction tuning on blocks.  also volt/duty-cycle.

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -31,39 +31,37 @@ public class Falcon6DriveMotor implements MotorWithEncoder100<Distance100> {
...
 
     /**
      * Friction feedforward in volts, for when the mechanism is moving.
      * 
      * This value seems very low, perhaps because the falcon closed-loop control is
      * compensating?
      */
-    private static final double dynamicFrictionFFVolts = 0.01;
+    private static final double dynamicFrictionFFVolts = 0.019;
```
truher, line 48: same, also volt/duty-cycle.

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -334,28 +332,28 @@ public void reset() {
     private static double frictionFF(double currentMotorRev_S, double desiredMotorRev_S) {
         double direction = Math.signum(desiredMotorRev_S);
         if (currentMotorRev_S < staticFrictionSpeedLimitRev_S) {
-            return staticFrictionFFVolts * direction / saturationVoltage;
+            return staticFrictionFFVolts * direction ;
```
truher, line 335: by removing saturation voltage you are changing the meaning of all the FF methods to return volts instead of duty cycle.  it does seem like using voltage feedforward would be better, but if you want to do that, you have to also change all the comments and change the motor.setControl() method to use VelocityVoltage instead of VelocityDutyCycle.

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
```diff
@@ -34,18 +34,18 @@ public class Falcon6TurningMotor implements Motor100<Angle100> {
...
      */
-    private static final double staticFrictionSpeedLimitRev_S = 1.0;
+    private static final double staticFrictionSpeedLimitRev_S = 0.1;
 
     /**
      * Friction feedforward in volts, for when the mechanism is stopped, or nearly
      * so.
      */
-    private static final double staticFrictionFFVolts = 0.375;
+    private static final double staticFrictionFFVolts = 0.375/20;
```
truher, line 43: don't do this sort of math in constants, just write what you think the value should be.

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
```diff
@@ -63,7 +63,7 @@ public class Falcon6TurningMotor implements Motor100<Angle100> {
      * Proportional feedback coefficient for the controller. The error is measured
      * in sensor units (ticks per 100ms), and the full scale output is 1023.
      */
-    private static final double outboardP = 0.1;
+    private static final double outboardP = 1/2048;
```
truher, line 66: change the comment to match the meaning you have in your head when you write this number.  error measured in ... rev/s?  also don't use an expression like this, just write a literal.

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
```diff
@@ -226,7 +227,7 @@ public void close() {
     @Override
     public void periodic() {
         // m_rawVelocity = m_motor.getSelectedSensorVelocity();
-        m_velocityRev_S = m_motor.getVelocity().getValueAsDouble();
+        m_velocityRev_S = m_motor.getVelocity().getValueAsDouble()/m_gearRatio;
```
truher, line 230: same gear ratio bug as above

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
```diff
@@ -240,28 +241,28 @@ public void periodic() {
 
     /** Velocity feedforward in duty cycle units [-1, 1] */
     private static double velocityFF(double desiredMotorRev_S) {
-        return velocityFFVoltS_Rev * desiredMotorRev_S / saturationVoltage;
+        return velocityFFVoltS_Rev * desiredMotorRev_S ;
```
truher, line 244: same duty cycle bug as above

studies/motordemo/vendordeps/Phoenix6.json
```diff
@@ -1,7 +1,7 @@
 {
     "fileName": "Phoenix6.json",
     "name": "CTRE-Phoenix (v6)",
-    "version": "24.0.0-beta-2",
+    "version": "24.0.0-beta-8",
```
truher, line 4: if you are somehow still on the beta version of something, you should upgrade to 2024.2.1.

truher (changes requested):
lots of issues here.

lib/src/main/java/org/team100/lib/config/NoteDetector.java
Coolearthsky: Fixed

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -60,10 +60,8 @@ public class Falcon6DriveMotor implements MotorWithEncoder100<Distance100> {
     private static final double accelFFVoltS2_M = 0;
 
     /**
-     * Proportional feedback coefficient for the controller. The error is measured
-     * in sensor units (ticks per 100ms), and the full scale output is 1023.
      */
-    private static final double outboardP = 0.05;
+    private static final double outboardP = .001;
```
truher, line 64: it doesn't really make sense that the P value should be less if the error value is also less.  it seems like you're just effectively turning off the feedback controller.   maybe add a TODO here to figure it out?

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -231,12 +218,13 @@ public void setVelocity(double outputM_S, double accelM_S_S) {
...
-        v.Acceleration = motorRev_S2;
         v.EnableFOC = true;
+        v.Acceleration = motorRev_S2;
         m_motor.setControl(v);
 
-        // m_motor.set(ControlMode.Velocity, motorTick_100ms, DemandType.ArbitraryFeedForward, kFF);
-
+        // m_motor.set(ControlMode.Velocity, motorTick_100ms,
+        // DemandType.ArbitraryFeedForward, kFF);
+        t.log(Level.DEBUG, m_name, "motor input", motorRev_S);
```
truher, line 227: put the unit in the name

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -334,28 +321,28 @@ public void reset() {
     private static double frictionFF(double currentMotorRev_S, double desiredMotorRev_S) {
         double direction = Math.signum(desiredMotorRev_S);
         if (currentMotorRev_S < staticFrictionSpeedLimitRev_S) {
-            return staticFrictionFFVolts * direction / saturationVoltage;
+            return staticFrictionFFVolts * direction;
```
truher, line 324: still looks like there's confusion about the unit here.  is it duty cycle or volts?

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -231,12 +218,13 @@ public void setVelocity(double outputM_S, double accelM_S_S) {
 
         VelocityDutyCycle v = new VelocityDutyCycle(motorRev_S);
```
truher, line 219: the drive motor uses VelocityDutyCycle but the turning motor uses VelocityVoltage,and really the best option is VelocityTorqueCurrentFOC.

truher (changes requested):

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
Coolearthsky: Changed to VelocityTorqueCurrentFOC

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
Coolearthsky: Done

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -334,28 +311,28 @@ public void reset() {
...
         if (currentMotorRev_S < staticFrictionSpeedLimitRev_S) {
-            return staticFrictionFFVolts * direction / saturationVoltage;
+            return staticFrictionFFAmps * direction;
         }
-        return dynamicFrictionFFVolts * direction / saturationVoltage;
+        return dynamicFrictionFFAmps * direction;
     }
 
     /**
      * Velocity feedforward in duty cycle units [-1, 1]
```
truher, line 320: the comments say duty cycle but the return value seems to be in amps?

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
```diff
@@ -240,28 +235,28 @@ public void periodic() {
 
     /** Velocity feedforward in duty cycle units [-1, 1] */
```
truher, line 236: same issue with duty cycle comments here

truher (changes requested):

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
```diff
@@ -26,49 +26,45 @@
...
+     * Friction feedforward in amps, for when the mechanism is moving.
      */
-    private static final double dynamicFrictionFFVolts = 0.27;
+    private static final double dynamicFrictionFFAmps = 0.27;
 
     /**
-     * Velocity feedforward in units of volts per motor revolution per second, or
-     * volt-seconds per revolution. Since saturation is 11 volts and free speed is
+     * Velocity feedforward in units of amps per motor revolution per second, or
+     * amp-seconds per revolution. Since saturation is 11 amp and free speed is
```
truher, line 52: you can't just search and replace the word "volt" without changing the meaning of this sentence.

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -26,54 +26,43 @@
...
      * compensating?
      */
-    private static final double dynamicFrictionFFVolts = 0.01;
+    private static final double dynamicFrictionFFAmps = 0.01;
 
     /**
-     * Velocity feedforward in units of volts per motor revolution per second, or
-     * volt-seconds per revolution. Since saturation is 11 volts and free speed is
+     * Velocity feedforward in units of amps per motor revolution per second, or
+     * amp-seconds per revolution. Since saturation is 11 amps and free speed is
```
truher, line 52: you can't just search and replace the word "volts" without changing the meaning of this sentence.

truher (changes requested):

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
```diff
@@ -26,49 +26,43 @@
...
+    private static final double velocityFFAmps_Rev = 0.11;
 
     /**
      * Placeholder for accel feedforward.
      */
-    private static final double accelFFVoltS2_Rad = 0;
+    private static final double accelFFAmps2_Rad = 0;
 
     /**
      * Proportional feedback coefficient for the controller. The error is measured
```
truher, line 61: this comment is now wrong

truher (changes requested):

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
Coolearthsky: Fixed

PR #203, opened 2024-02-10
Pheonix 6 code and more

Coolearthsky:


lib/src/main/java/org/team100/lib/motion/drivetrain/kinodynamics/SwerveKinodynamicsFactory.java
```diff
@@ -35,7 +35,7 @@ public static SwerveKinodynamics get() {
                 // TODO: make tests specify kinodynamics instead.
                 return new SwerveKinodynamics(4, 2, 3, 13, 20 * Math.PI, 0.5, 0.5, 0.3);
             case BETA_BOT:
-                return new SwerveKinodynamics(4, 2, 3, 13, 20 * Math.PI, 0.491, 0.765, 0.3);
+                return new SwerveKinodynamics(5, 5, 4, 13, 20 * Math.PI, 0.491, 0.765, 0.3);
```
truher, line 38: the third number in this list is max deceleration which should certainly be higher than max acceleration.

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -37,20 +37,20 @@ public class Falcon6DriveMotor implements MotorWithEncoder100<Distance100> {
...
+    private static final double staticFrictionFFAmps = 7;
 
     /**
      * Friction feedforward in amps, for when the mechanism is moving.
      * 
      * This value seems very low, perhaps because the falcon closed-loop control is
      * compensating?
      */
-    private static final double dynamicFrictionFFAmps = 0.01;
+    private static final double dynamicFrictionFFAmps = 5;
```
truher, line 48: remove the comment about the value being low

lib/src/main/java/org/team100/lib/motor/drive/Falcon6DriveMotor.java
```diff
@@ -212,6 +211,7 @@ public void setVelocity(double outputM_S, double accelM_S_S) {
 
         // m_motor.set(ControlMode.Velocity, motorTick_100ms,
         // DemandType.ArbitraryFeedForward, kFF);
+        t.log(Level.DEBUG, m_name, m_name, wheelRev_S);
```
truher, line 214: use a descriptive name here

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
```diff
@@ -40,17 +40,17 @@ public class Falcon6TurningMotor implements Motor100<Angle100> {
      * Friction feedforward in amps, for when the mechanism is stopped, or nearly
      * so.
      */
-    private static final double staticFrictionFFAmps = 0.375;
+    private static final double staticFrictionFFAmps = 0;
```
truher, line 43: is this a mistake?

truher (changes requested):

lib/src/main/java/org/team100/lib/motor/turning/Falcon6TurningMotor.java
Coolearthsky: This is what worked in the testing, I am guessing that since the FF was so hard to tune the PID is just doing most of the work, and we likely will still need to tune the feedfoward.

truher: ok, add a TODO

Coolearthsky: done

Coolearthsky: PTAL

Coolearthsky: oh also I moved the feedforward up so the low level feedforward class, so the tuning TODO is there now

PR #218, opened 2024-02-16
Moved up camera angle to python code

Coolearthsky:


lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -52,61 +33,56 @@ public CameraAngles(
...
         m_yOffset = 0;
     }
 
     /**
+     *         0 should be center of FOV
      * @return A robot relative translational x value of an object in a camera in
      *         meters
      */
-    public double getX(double vertPixels) {
+    public double getX(double vertFOVRads) {
```
truher, line 47: i think you mean this argument to be the rotation to the target, it doesn't have anything to do with "FOV" so why is it called that?  i think you should clarify what this represents.

lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -52,61 +33,56 @@ public CameraAngles(
...
      *         meters
      */
-    public double getY(double horzPixels, double vertPixels) {
-        double x = getX(vertPixels);
-        double y = -1.0 * x
-                * Math.tan(Math.toRadians(m_horzFOVDegrees) * (horzPixels
-                        - m_horzResolution / 2) / m_horzResolution);
+    public double getY(double vertFOVRads, double horzFOVRads) {
+        double x = -1.0 * getX(vertFOVRads);
+        double y = x * Math.tan(horzFOVRads + m_horzAngle);
```
truher, line 60: same question about offset here.  aren't camera offsets "position of camera relative to robot" i.e. if the camera is on the left side then it has a positive offset?

comp/vision/gamepiece_finder24.py
```diff
@@ -191,13 +197,14 @@ def find_object(self, img_yuv):
 
             cX = int(mmnts["m10"] / mmnts["m00"])
             cY = int(mmnts["m01"] / mmnts["m00"])
-
+            xAngle = ((cY - (self.height/2))/self.height)*self.vertFOV
+            yAngle = ((cX - (self.width/2))/self.width)*self.horzFOV
```
truher, line 201: i think here you're trying to represent "yaw" which is positive-right in camera coordinates

lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -52,61 +33,56 @@ public CameraAngles(
...
+     *         0 should be center of FOV 
      * @return A robot relative translational y value of an object in a camera in
      *         meters
      */
-    public double getY(double horzPixels, double vertPixels) {
-        double x = getX(vertPixels);
-        double y = -1.0 * x
-                * Math.tan(Math.toRadians(m_horzFOVDegrees) * (horzPixels
-                        - m_horzResolution / 2) / m_horzResolution);
+    public double getY(double vertFOVRads, double horzFOVRads) {
```
truher, line 58: same issue with the word "FOV" here.  i'm pretty sure you're trying to represent the same concept that WPILib uses Rotation3d for, so maybe you should just do that?

also it would be good for x and y not to be two separate public methods.  make one public method that takes a camera-coordinate thing (perhaps Rotation3d) and returns a robot-coordinate thing (perhaps Translation2d)

lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -1,47 +1,28 @@
 package org.team100.lib.util;
 
 public class CameraAngles {
-    private final double m_downwardAngleDegrees;
-    private final double m_horzFOVDegrees;
-    private final double m_vertFOVDegrees;
-    private final double m_horzResolution;
-    private final double m_vertResolution;
+    private final double m_downwardAngleRads;
+    private final double m_horzAngle = 0;
```
truher, line 5: seems like this isn't used?

we have a way to remember camera offsets (see the Camera class), so you should use that instead of making another one here.

lib/src/test/java/org/team100/lib/localization/TestCameraAngles.java
```diff
@@ -10,11 +10,12 @@ class TestCameraAngles extends Fixtured {
     @Test
     void testCameraAngles() {
         {
-            CameraAngles camera = new CameraAngles(30, 67.5, 50, 832, 616, 1, 0,0);
-               assertEquals(0,camera.getY(416,200),0.0001);
-               assertEquals(-0.7,camera.getX(0),0.001);
-               assertEquals(416,camera.getInverseY(1,0),0.001);
-               assertEquals(123.2,camera.getInverseX(1),0.001);
+            CameraAngles camera = new CameraAngles(Math.toRadians(30), 1, 0,0);
```
truher, line 13: i think you should make a few more very simple test cases here to be sure you have the transformation logic correct.  like if the camera returns a target in the center (i.e. yaw = 0, pitch = 0) and if the camera offset is something very easy to calculate like 1 meter up, 45 degrees down, do you get the correct result (1 meter ahead)?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -57,12 +64,12 @@ void consumeValues(NetworkTableEvent e) {
      * @return The x position in the camera in pixels, 0 should be the left of the
      *         screen
      */
-    public Optional<Double> getX() {
+    private Optional<Double> getX() {
```
truher, line 67: remove the "positions" field and just pass what you need to this function.  it is generally a bad idea for one method to stash something in a field so that another method can pick it up.  just pass it as an argument.

also what does this represent?  the comment says it's pixels.  but it's used below as some sort of yaw measurement.

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -1,41 +1,37 @@
...
 import edu.wpi.first.math.geometry.Transform2d;
 import edu.wpi.first.math.geometry.Translation2d;
 
-public class NoteDetector {
-    private final CameraAngles m_cameraAngles;
-    private final SwerveDriveSubsystem m_swerve;
-    private final NotePosition24ArrayListener m_notePosition24ArrayListener;
+public class NotePoseDetector {
+    private final SwerveDrivePoseEstimator100 m_swervePose;
+    private Translation2d m_robotRelativeNoteTranslation = new Translation2d();
```
truher, line 12: don't use a field here.  it is generally a bad idea for one method to stash something in a field so that some other method picks it up.  just pass it as an argument.

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -134,7 +130,7 @@ public Pose2d fieldRelativePose2d() {
         switch (Identity.instance) {
             case BETA_BOT:
             case COMP_BOT:
-                return m_swerve.getPose().transformBy(robotRelativeTransform2d());
+                return m_swervePose.getEstimatedPosition().transformBy(robotRelativeTransform2d());
```
truher, line 133: i'm confused about what the "rotation" of this object represents.  the camera isn't sensing rotation.  what is being measured by the rotational component of this pose?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -73,18 +80,25 @@ public Optional<Double> getX() {
      * @return The y position in the camera in pixels, 0 should be the bottom of the
      *         screen
      */
-    public Optional<Double> getY() {
+    private Optional<Double> getY() {
```
truher, line 83: what does this represent?  the comment says something about pixels.  but it's used below as pitch.  what is this?

lib/src/test/java/org/team100/lib/localization/TestCameraAngles.java
```diff
@@ -4,17 +4,36 @@
...
+               assertEquals(0,testTranslation1.getY(),0.0001);
+               assertEquals(0.7,testTranslation2.getX(),0.001);
+               assertEquals(0.123465481082,testTranslation2.getY(),0.001);
+            //    assertEquals(416,camera.getInverseY(1,0),0.001);
+            //    assertEquals(123.2,camera.getInverseX(1),0.001);
+            }
+    }
+
+    @Test
+    void differentCameraAngles() {
```
truher, line 31: add a few more easy-to-calculate cases.

truher (changes requested):

lib/src/test/java/org/team100/lib/localization/TestCameraAngles.java
Coolearthsky: Added

lib/src/main/java/org/team100/lib/localization/NotePosition24.java
```diff
@@ -3,21 +3,25 @@
 public class NotePosition24 {
     private final int x;
     private final int y;
+
+    // Yaw and Pitch angles to object in camera
```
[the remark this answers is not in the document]
Coolearthsky: fixed

lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -1,112 +1,49 @@
...
+     * 
      * @return A robot relative translational x value of an object in a camera in
      *         meters
      */
-    public double getX(double vertPixels) {
-        double x = -1.0 * m_cameraHeightMeters
-                * Math.tan(vertPixels * (Math.toRadians(m_vertFOVDegrees / m_vertResolution))
-                        + Math.toRadians(90 - m_vertFOVDegrees / 2 - m_downwardAngleDegrees));
-        return x - m_xOffset;
+    private double getX(double vertRotToTargetRads) {
```
truher, line 32: inline this method

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -43,48 +48,83 @@ void consumeValues(NetworkTableEvent e) {
...
+     * @return The yaw to the object in the camera, 0 is center
      */
     public Optional<Double> getX() {
         switch (Identity.instance) {
             case BETA_BOT:
-                double xd = positions[0].getX();
-                Double x = xd;
-                Optional<Double> e = Optional.of(x);
+            case COMP_BOT:
+                double xd = translation.getX();
```
truher, line 106: why is this inside a switch statement?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -43,48 +48,83 @@ void consumeValues(NetworkTableEvent e) {
...
+                return e;
+            default:
+                return Optional.empty();
+        }
+    }
+
+    private void updateMeasurement(String ID) {
+        Transform3d cameraInRobotCoordinates = Camera.get(ID).getOffset();
+        CameraAngles e = new CameraAngles(cameraInRobotCoordinates);
+        double dx = positions[0].getYaw();
```
truher, line 82: looks like you're just using the first entry in the position array.  why?  does the python pick the best target as the first entry?

Coolearthsky: To be honest, I was going to do better testing on this, I don't exactly know how to but I was planning on picking either the biggest contour, i'll add a todo to take care of it tomorrow

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -40,51 +46,74 @@ void consumeValues(NetworkTableEvent e) {
...
             System.out.println("note weird vision update key: " + name);
         }
     }
 
-    /**
-     * @return The x position in the camera in pixels, 0 should be the left of the
-     *         screen
-     */
-    public Optional<Double> getX() {
+    private Optional<Double> optionalX(double position) {
```
truher, line 58: what is this method for?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -16,14 +20,16 @@
 public class NotePosition24ArrayListener {
 
     StructBuffer<NotePosition24> m_buf = StructBuffer.create(NotePosition24.struct);
-    NotePosition24[] positions = new NotePosition24[0];
+    NotePosition24[] positions;
+    private String id;
```
truher, line 24: why is this a field of this class?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -40,51 +46,74 @@ void consumeValues(NetworkTableEvent e) {
                     positions = m_buf.readArray(b);
                     latestTime = Timer.getFPGATimestamp();
                 }
+              positions
```
truher, line 49: typo here?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -40,51 +46,74 @@ void consumeValues(NetworkTableEvent e) {
...
+            case BETA_BOT:
+            case COMP_BOT:
+                Optional<Double> e = Optional.of(position);
+                return e;
+            default:
+                return Optional.empty();
+        }
+    }
+
+    private Translation2d getTranslationNonOptional() {
```
truher, line 80: what is this method for?

truher (changes requested):

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile.java
```diff
@@ -66,36 +123,40 @@ public void initialize100() {
...
-        SwerveState goalState = new SwerveState(xSetpoint, ySetpoint, thetaSetpoint);
-        Twist2d goal = m_controller.calculate(m_swerve.getState(), goalState);
-        m_swerve.driveInFieldCoords(goal, 0.02);
+            // make sure the setpoint uses the modulus close to the measurement.
+            thetaSetpoint = new State100(
+                    Math100.getMinDistance(measurement, thetaSetpoint.x()),
+                    thetaSetpoint.v());
+            State100 thetaGoal = new State100(bearing.getRadians(), 0);
+            State100 xGoalRaw = new State100(m_fieldRelativeGoal.get().getX(), 0, 0);
+            xSetpoint = xProfile.calculate(0.02, xSetpoint, xGoalRaw);
```
truher, line 138: replace all these instances of "0.02" with "dt" since it's a better estimate of the duration of the following period

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile.java
```diff
@@ -66,36 +123,40 @@ public void initialize100() {
...
             m_swerve.driveInFieldCoords(new Twist2d(), 0.02);
-        }    
+        }
     }
 
     @Override
     public boolean isFinished() {
-        //TODO make this end when intake detects note intake
-        return false;
+        if (m_end != null) {
```
truher, line 155: it's better to just make m_end never be null: your constructors can guarantee it has a value.

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile.java
```diff
@@ -16,46 +17,102 @@
...
-    // inject these, make them the same as the kinematic limits, inside the
-    // trajectory supplier.
     private final Supplier<Pose2d> m_fieldRelativeGoal;
     private final SwerveDriveSubsystem m_swerve;
     private final HolonomicDriveController100 m_controller;
     private final SwerveKinodynamics m_limits;
     private final TrapezoidProfile100 xProfile;
     private final TrapezoidProfile100 yProfile;
     private final TrapezoidProfile100 thetaProfile;
+    private BooleanSupplier m_end;
```
truher, line 28: make m_end final

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile.java
```diff
@@ -16,46 +17,102 @@
...
+    private boolean m_endBoolean;
```
truher, line 29: remove this.  why do you have two kinds?

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -81,10 +77,9 @@ public Rotation2d fieldRelativeAngleToNote() {
      * @return A robot relative translational y value of an object in a camera in
      *         meters
      */
-    public Double getY() {
-        return m_cameraAngles.getY(m_notePosition24ArrayListener.getX().get(),
-                m_notePosition24ArrayListener.getY().get());
-    }
+    public double getY() {
```
truher, line 80: remove this.  callers can just get the Translation2d and pick out the x and y if they want to.

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -34,8 +28,8 @@ public NoteDetector(
      * @return A robot relative translational x value of an object in a camera in
      *         meters
      */
-    public double getX() {
-        return m_cameraAngles.getX(m_notePosition24ArrayListener.getY().get());
+    public Double getX() {
```
truher, line 31: remove this.  callers can just get the Translation2d and pick out the x and y if they want to.

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -46,9 +40,9 @@ public Double fieldRelativeX() {
         switch (Identity.instance) {
             case BETA_BOT:
             case COMP_BOT:
-                return fieldRelativePose2d().getX();
             case BLANK:
-                return 0.0;
+                return fieldRelativePose2d().getX();
```
truher, line 44: there's no reason for the fieldRelativeX method to exist

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -69,9 +65,9 @@ public Rotation2d fieldRelativeAngleToNote() {
         switch (Identity.instance) {
             case BETA_BOT:
             case COMP_BOT:
+            case BLANK:
```
truher, line 68: it looks to me like there's an infinite loop here: fieldRelativeAngleToNote calls fieldRelativePose2d which callsFieldRelativeAngleToNote.

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -123,7 +121,7 @@ public Translation2d FieldRelativeTranslation2d() {
      *         for translation and radians for rotation
      */
     public Transform2d robotRelativeTransform2d() {
-        return new Transform2d(getX(), getY(), robotRelativeAngleToNote());
+        return new Transform2d(RobotRelativeTranslation2d(), robotRelativeAngleToNote());
```
truher, line 124: this transform involves a translation to the note and also a rotation to face it along the line from the robot position to the note position.  since these two things, the translation and the rotation, actually describe the same thing, just let the caller decide how they want to rotate.

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -107,7 +102,10 @@ public Double fieldRelativeY() {
      *         meters, rot value is bearing
      */
     public Translation2d RobotRelativeTranslation2d() {
-        return new Translation2d(getX(), getY());
+        if (m_notePosition24ArrayListener.getTranslation2d()[0].isPresent()) {
+        return m_notePosition24ArrayListener.getTranslation2d()[0].get();
+        }
+        return new Translation2d();
```
truher, line 108: if there's no value, return Optional.empty().  if you return a real value here, then you invite the caller to interpret it as a real value.

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -54,38 +58,33 @@ void consumeValues(NetworkTableEvent e) {
...
+    */
+    public Optional<Translation2d>[] getTranslation2d() {
+        Optional<Translation2d>[] notes = new Optional[]{Optional.empty()};
+        int noteNum = 0;
+        Optional<Double> y = Optional.empty();
+        Optional<Double> x = Optional.empty();
+        Transform3d cameraInRobotCoordinates = Camera.get(id).getOffset();
+        CameraAngles ed = new CameraAngles(cameraInRobotCoordinates);
+        for (NotePosition24 position : positions) {
+            if (position.getPitch() < Math.PI / 2 - cameraInRobotCoordinates.getRotation().getY()) {
```
truher, line 71: i'm confused about this logic.  isn't the pitch measured from straight ahead?  how does this inequality represent the horizon?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
Coolearthsky: Would it be better to have it as a CameraAngles field and just use the ID to feed the parems?

lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -1,112 +1,29 @@
...
-        return y;
-    }
-
-    /**
-     * @return A robot relative angle in radians to the note
-     */
-    public double getAngle(double horzPixels, double vertPixels) {
-        return Math.atan2(getY(horzPixels, vertPixels), getX(vertPixels));
+    public Translation2d getTranslation2d(Rotation3d cameraObject) {
+        double x = m_cameraInRobotCoordinates.getZ() * Math.tan(cameraObject.getY() + m_cameraInRobotCoordinates.getRotation().getY());
```
truher, line 25: i'm confused by this logic.  if the camera offset is zero, then it's facing straight ahead.  if the camera "object" is at zero (i.e. in the center) then it should be at the horizon, infinitely far away.  but this logic would return an x value of zero.  try writing a few more tests to verify this logic.

truher (changes requested):

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
truher: i think you are imagining a specific sequence of method calls in this class, but nothing guarantees that sequence.  the "consumeValues" method can be called at any time by a dedicated networktables listener thread.  the other methods can be called at any time by a different thread.  you should do *all* the work of calculation in the consumeValues method, and the other methods should just retrieve the results atomically.

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
Coolearthsky: It would be facing downward if the camera pitch object is zero

lib/src/main/java/org/team100/lib/util/CameraAngles.java
Coolearthsky: It would be facing downward if the camera pitch object is zero

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
truher: in what coordinates?  what is the pitch measuring?

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
truher: I guess.  the main issue is just way too much code here, you can remove most of these methods.

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -0,0 +1,97 @@
...
+            default:
+                return Optional.empty();
+        }
+    }
+
+    /**
+     * @return A field relative Pose2d value of an object in a camera in meters for
+     *         translation and radians for rotation
+     */
+    public Optional<Pose2d> fieldRelativePose2d(Rotation2d angle) {
```
truher, line 62: what does this angle mean?  is it the rotation of the note?  the note is rotationally symmetric; is rotation important here?

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -0,0 +1,97 @@
...
+            SwerveDriveSubsystem swerve) {
+        m_notePosition24ArrayListener = notePosition24ArrayListener;
+        m_swerve = swerve;
+    }
+
+    /**
+     * @return A robot relative translational value of an object in a camera in
+     *         meters, rot value is bearing
+     */
+    public Optional<Translation2d> RobotRelativeTranslation2d() {
```
truher, line 33: this method doesn't seem to do anything; if you just return the translation, you'll be returning optional.empty if it's not present.

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -43,8 +48,22 @@ void consumeValues(NetworkTableEvent e) {
...
             }
-
+            notes = new Optional[] { Optional.empty() };
+            int noteNum = 0;
+            Optional<Double> y = Optional.empty();
+            Optional<Double> x = Optional.empty();
+            Transform3d cameraInRobotCoordinates = Camera.get(fields[1]).getOffset();
+            CameraAngles ed = new CameraAngles(cameraInRobotCoordinates);
             for (NotePosition24 position : positions) {
+                if (position.getPitch() < Math.PI / 2 - cameraInRobotCoordinates.getRotation().getY()) {
```
truher, line 58: i'd suggest extracting this little bit of geometry to a utility class like PoseEstimatorHelper or GeometryUtil, so you can write some tests to verify it.

lib/src/main/java/org/team100/lib/config/NotePoseDetector.java
```diff
@@ -0,0 +1,97 @@
...
+
+import org.team100.lib.localization.NotePosition24ArrayListener;
+import org.team100.lib.motion.drivetrain.SwerveDriveSubsystem;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.geometry.Transform2d;
+import edu.wpi.first.math.geometry.Translation2d;
+
+public class NotePoseDetector {
```
truher, line 13: it's not clear to me that this class needs to exist at all.  it seems like it just does a tiny bit of geometry, which is only used by the command?  maybe extract the geometry as a static method in a utility class like PoseEstimatorHelper and get rid of this class

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -14,18 +18,18 @@
 
 /** For testing the NotePosition struct array */
 public class NotePosition24ArrayListener {
-
     StructBuffer<NotePosition24> m_buf = StructBuffer.create(NotePosition24.struct);
-    NotePosition24[] positions = new NotePosition24[0];
+    Optional<Translation2d>[] notes = new Optional[] { Optional.empty() };;
```
truher, line 22: why does this array have more than nothing in it?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -43,8 +48,22 @@ void consumeValues(NetworkTableEvent e) {
             } catch (RuntimeException ex) {
                 return;
             }
-
+            notes = new Optional[] { Optional.empty() };
+            int noteNum = 0;
+            Optional<Double> y = Optional.empty();
```
truher, line 53: why are these declarations here, when the values are only ever used inside the loop?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -43,8 +48,22 @@ void consumeValues(NetworkTableEvent e) {
...
+            Optional<Double> x = Optional.empty();
+            Transform3d cameraInRobotCoordinates = Camera.get(fields[1]).getOffset();
+            CameraAngles ed = new CameraAngles(cameraInRobotCoordinates);
             for (NotePosition24 position : positions) {
+                if (position.getPitch() < Math.PI / 2 - cameraInRobotCoordinates.getRotation().getY()) {
+                    double dy = positions[0].getPitch();
+                    Double yz = dy;
+                    y = Optional.of(yz);
```
truher, line 61: why is this optional, when it is only used once, a few lines down?

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -43,8 +48,22 @@ void consumeValues(NetworkTableEvent e) {
...
-
+            notes = new Optional[] { Optional.empty() };
+            int noteNum = 0;
+            Optional<Double> y = Optional.empty();
+            Optional<Double> x = Optional.empty();
+            Transform3d cameraInRobotCoordinates = Camera.get(fields[1]).getOffset();
+            CameraAngles ed = new CameraAngles(cameraInRobotCoordinates);
             for (NotePosition24 position : positions) {
+                if (position.getPitch() < Math.PI / 2 - cameraInRobotCoordinates.getRotation().getY()) {
+                    double dy = positions[0].getPitch();
```
truher, line 59: you shouldn't use variable names like "y" when the real meaning is "pitch."

lib/src/main/java/org/team100/lib/util/CameraAngles.java
```diff
@@ -1,112 +1,29 @@
...
-        double y = m_horzResolution * (Math.atan((sideMeters+m_yOffset)/(forwardMeters+m_xOffset))/Math.toRadians(m_horzFOVDegrees)) + m_horzResolution/2;
-        return y;
-    }
-
-    /**
-     * @return A robot relative angle in radians to the note
-     */
-    public double getAngle(double horzPixels, double vertPixels) {
-        return Math.atan2(getY(horzPixels, vertPixels), getX(vertPixels));
+    public Translation2d getTranslation2d(Rotation3d cameraObject) {
```
truher, line 24: wpi supplies methods for composing 3d geometries, and you should use them.  so if you have camera in robot coordinates, and a rotation in the camera coordinates, then to get the rotation in robot coordinates you just say cameraInRobotCoordinates.plus(new Transform3d(new Translation3d(), rotationInCameraCoordinates);

also i think it would be more clear if you made the calculation more physically meaningful.  you're calculating the complement of the tilt angle, i.e. the angle between the vertical and the note, so just do that.  as it's written, it seems like you're calculating a different (large) angle that happens to have the same tangent as the one you really want.  it would be good to draw a picture of what you're doing.

also this can just be a static method in PoseEstimationHelper,, you don't need this class at all.

PR #252, opened 2024-02-25
CAN IMU classes and LED plate support

manouserres:


lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
+package frc.robot;
+
+import edu.wpi.first.hal.CANData;
+import edu.wpi.first.util.sendable.Sendable;
+import edu.wpi.first.util.sendable.SendableBuilder;
+import edu.wpi.first.wpilibj.CAN;
+
+public class Team100CANIMU {
+    public static final int CAN_DEVICE_MANUFACTURER = 8;
```
truher, line 9: i think these can all be private unless you need the callers to know about them

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+    public static final int API_ID_EULER_ANGLES = 17;
+    public static final int API_ID_QUATERNION = 18;
+    // the 16 bit values on the CAN bus must be divided by 163.84 then 200 should be
+    // subtracted from the value
+    public static final double INT_TO_ENG_COEFF = 1 / 163.84;
+    public static final double INT_TO_ENG_OFFSET = -200;
+    public static final double INT_TO_ENG_ACC = 1 / 655.36;
+
+    private long rollValueInt, pitchValueInt, yawValueInt, accValueInt;
+    private int deviceID;
```
truher, line 20: these can be final unless you need to change them during the lifespan of the Team100CANIMU object.

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+    private boolean result;
+    private CAN Team100CANIMU;
```
truher, line 22: change the name of this field so it's not the same as the class name

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+        EAngles.setPitch(pitchValueInt * INT_TO_ENG_COEFF + INT_TO_ENG_OFFSET);
+        yawValueInt = ((IMUData.data[4]<<8) & 0xFF00) + (IMUData.data[5] & 0xFF);
+        EAngles.setYaw(yawValueInt * INT_TO_ENG_COEFF + INT_TO_ENG_OFFSET);
+        accValueInt = ((IMUData.data[6]<<8) & 0xFF00) + (IMUData.data[7] & 0xFF);
+        EAngles.setAcc(accValueInt * INT_TO_ENG_ACC);
+        return EAngles;
+    }
+
+    /*
+     * Get Roll Angle
```
truher, line 61: you don't need these comments since they're the same as the method names

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+    public static final int API_ID_QUATERNION = 18;
+    // the 16 bit values on the CAN bus must be divided by 163.84 then 200 should be
+    // subtracted from the value
+    public static final double INT_TO_ENG_COEFF = 1 / 163.84;
+    public static final double INT_TO_ENG_OFFSET = -200;
+    public static final double INT_TO_ENG_ACC = 1 / 655.36;
+
+    private long rollValueInt, pitchValueInt, yawValueInt, accValueInt;
+    private int deviceID;
+    private boolean result;
```
truher, line 21: it doesn't look like "result" is ever actually used here (i.e. it is only ever written to).  did you mean to use it for something?

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+    public static final int CAN_DEVICE_TYPE = 4;
+    public static final int API_ID_EULER_ANGLES = 17;
+    public static final int API_ID_QUATERNION = 18;
+    // the 16 bit values on the CAN bus must be divided by 163.84 then 200 should be
+    // subtracted from the value
+    public static final double INT_TO_ENG_COEFF = 1 / 163.84;
+    public static final double INT_TO_ENG_OFFSET = -200;
+    public static final double INT_TO_ENG_ACC = 1 / 655.36;
+
+    private long rollValueInt, pitchValueInt, yawValueInt, accValueInt;
```
truher, line 19: i think these can all be local variables in the methods where they're used

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+     * Destructor
+     */
+//  public void close() {
+//    Team100CANIMU.close();
+//  }
+
+    /*
+     * Get Euler Angles
+     */
+    public EulerAngles getEulerAngles() {
```
truher, line 47: i usually include the unit in the method name, to avoid confusion, so this would be something like getEulerAnglesRad().

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+    private boolean result;
+    private CAN Team100CANIMU;
+    private CANData IMUData;
+    private EulerAngles EAngles;
+
+    /*
+     * Constructor
+     */
+    public Team100CANIMU(int deviceID) {
+        this.deviceID = deviceID;
```
truher, line 30: i don't think deviceid needs to be a field of the class, it's only passed to the CAN constructor.

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+//  public void close() {
+//    Team100CANIMU.close();
+//  }
+
+    /*
+     * Get Euler Angles
+     */
+    public EulerAngles getEulerAngles() {
+        result = Team100CANIMU.readPacketLatest(API_ID_EULER_ANGLES, IMUData);
+        rollValueInt = ((IMUData.data[0]<<8) & 0xFF00) + (IMUData.data[1] & 0xFF);
```
truher, line 49: i'd suggest pulling all the two-bytes-to-int code into a single function since it's always the same.  give it an offset and return an int.

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+    // subtracted from the value
+    public static final double INT_TO_ENG_COEFF = 1 / 163.84;
+    public static final double INT_TO_ENG_OFFSET = -200;
+    public static final double INT_TO_ENG_ACC = 1 / 655.36;
+
+    private long rollValueInt, pitchValueInt, yawValueInt, accValueInt;
+    private int deviceID;
+    private boolean result;
+    private CAN Team100CANIMU;
+    private CANData IMUData;
```
truher, line 23: i think this doesn't need to be a field since it never relevant outside the method it's used in.  the java memory management system is pretty good at local variables, it's ok to create a new CANData in every method call.

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+    private EulerAngles EAngles;
```
truher, line 24: same story here, just make this local

lib/src/main/java/org/team100/lib/sensors/Team100CANIMU.java
```diff
@@ -0,0 +1,88 @@
...
+import edu.wpi.first.wpilibj.CAN;
+
+public class Team100CANIMU {
+    public static final int CAN_DEVICE_MANUFACTURER = 8;
+    public static final int CAN_DEVICE_TYPE = 4;
+    public static final int API_ID_EULER_ANGLES = 17;
+    public static final int API_ID_QUATERNION = 18;
+    // the 16 bit values on the CAN bus must be divided by 163.84 then 200 should be
+    // subtracted from the value
+    public static final double INT_TO_ENG_COEFF = 1 / 163.84;
```
truher, line 15: what's "ENG" mean here?

truher (changes requested):
looks good!  i made several comments aimed at simplifying.  happy to chat about them anytime.

PR #271, opened 2024-02-27
Created simulated camera for game piece following

Coolearthsky:


truher (changes requested):
can you make the simulated camera a class with some unit tests?

lib/src/main/java/org/team100/lib/config/SimulatedCamera.java
```diff
@@ -0,0 +1,58 @@
...
+                        - m_cameraInRobotCoordinates.getRotation().getY();
+            } else {
+                pitch = Math.atan2(m_cameraInRobotCoordinates.getZ(), x)
+                        - m_cameraInRobotCoordinates.getRotation().getY();
+            }
+            double y = relative.getY() - m_cameraInRobotCoordinates.getY();
+            double yaw = MathUtil
+                    .angleModulus(Math.atan2(y, x) - m_cameraInRobotCoordinates.getRotation().getZ());
+            Rotation3d rot = new Rotation3d(0, pitch, yaw);
+            if (Math.abs(pitch) < Math.toRadians(31.5) && Math.abs(yaw) < Math.toRadians(40)) {
```
truher, line 42: what are these numbers, 31.5 and 40?

lib/src/main/java/org/team100/lib/config/SimulatedCamera.java
```diff
@@ -0,0 +1,58 @@
...
+    //TODO make this work with any camera yaw
+    public Optional<ArrayList<Translation2d>> findNotes(Pose2d robotPose, Translation2d[] notes) {
+        Optional<ArrayList<Translation2d>> optionalList = Optional.empty();
+        ArrayList<Translation2d> list = new ArrayList<>();
+        for (Translation2d note : notes) {
+            Pose2d pose = new Pose2d(note, new Rotation2d());
+            Translation2d relative = pose.relativeTo(robotPose).getTranslation();
+            double pitch;
+            double x = relative.getX() - m_cameraInRobotCoordinates.getX();
+            if (m_cameraInRobotCoordinates.getRotation().getZ() == Math.PI) {
```
truher, line 31: seems like a special yaw case of some kind?  why not just do the correct thing?

lib/src/main/java/org/team100/lib/config/SimulatedCamera.java
```diff
@@ -0,0 +1,58 @@
...
+
+    //TODO make this work with any camera yaw
+    public Optional<ArrayList<Translation2d>> findNotes(Pose2d robotPose, Translation2d[] notes) {
+        Optional<ArrayList<Translation2d>> optionalList = Optional.empty();
+        ArrayList<Translation2d> list = new ArrayList<>();
+        for (Translation2d note : notes) {
+            Pose2d pose = new Pose2d(note, new Rotation2d());
+            Translation2d relative = pose.relativeTo(robotPose).getTranslation();
+            double pitch;
+            double x = relative.getX() - m_cameraInRobotCoordinates.getX();
```
truher, line 30: there are 3d transforms that make all this logic much simpler

lib/src/main/java/org/team100/lib/localization/NotePosition24ArrayListener.java
```diff
@@ -84,27 +89,28 @@ void consumeValues(NetworkTableEvent e) {
      * @return The translation of all the notes, field relative
      */
     public Optional<ArrayList<Translation2d>> getTranslation2dArray() {
-        if (latestTime > Timer.getFPGATimestamp() - 0.1) {
-            return notes;
+        switch (Identity.instance) {
+            case BLANK:
+                Transform3d cameraInRobotCoordinates = Camera.get("10000000e31d4a24").getOffset();
```
truher, line 94: why is this key here?  why not just say Camera.A or whatever?

lib/src/main/java/org/team100/lib/config/SimulatedCamera.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class SimulatedCamera {
+    private final Transform3d m_cameraInRobotCoordinates;
+
+    public SimulatedCamera(Transform3d cameraInRobotCoordinates) {
+        m_cameraInRobotCoordinates = cameraInRobotCoordinates;
+    }
+
+    //TODO make this work with any camera yaw
+    public Optional<ArrayList<Translation2d>> findNotes(Pose2d robotPose, Translation2d[] notes) {
```
truher, line 23: note this method isn't really "finding" anything, it's just transforming one planar geometry into another, so maybe call it something that reflects what it's actually doing?

lib/src/main/java/org/team100/lib/config/SimulatedCamera.java
```diff
@@ -0,0 +1,58 @@
...
+            Rotation3d rot = new Rotation3d(0, pitch, yaw);
+            if (Math.abs(pitch) < Math.toRadians(31.5) && Math.abs(yaw) < Math.toRadians(40)) {
+                Translation2d cameraRotationToRobotRelative = PoseEstimationHelper
+                        .cameraRotationToRobotRelative(
+                                m_cameraInRobotCoordinates,
+                                rot);
+                Translation2d l = PoseEstimationHelper.convertToFieldRelative(
+                        robotPose,
+                        cameraRotationToRobotRelative);
+                if (Math.abs(l.minus(note).getX()) < 0.01 && Math.abs(l.minus(note).getY()) < 0.01) {
```
truher, line 50: what is this logic?  it seems like maybe you're combining the logic of the camera with the logic of the detector?

lib/src/main/java/org/team100/lib/config/SimulatedCamera.java
```diff
@@ -0,0 +1,58 @@
...
+import org.team100.lib.localization.PoseEstimationHelper;
+
+import edu.wpi.first.math.MathUtil;
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.geometry.Rotation3d;
+import edu.wpi.first.math.geometry.Transform3d;
+import edu.wpi.first.math.geometry.Translation2d;
+
+public class SimulatedCamera {
```
truher, line 15: i'm confused about what this class does.  it seems like it's partly the camera transformation (which could be literally one line) and it's partly some other logic.  what does this class **MEAN**?

lib/src/main/java/org/team100/lib/localization/PoseEstimationHelper.java
```diff
@@ -27,17 +27,20 @@ public class PoseEstimationHelper {
     private static final Telemetry t = Telemetry.get();
     private static final String m_name = Names.name(PoseEstimationHelper.class);
 
-    public static Translation2d cameraRotationToRobotRelative(Transform3d cameraInRobotCoordinates, Rotation3d cameraObject) {
+    public static Translation2d cameraRotationToRobotRelative(Transform3d cameraInRobotCoordinates, Rotation3d yawPitch) {
+            Translation2d roll = new Translation2d(yawPitch.getZ(),yawPitch.getY());
```
truher, line 31: so you have a **TRANSLATION** called **ROLL**.   which is made from a variable that has the worlds **YAW** and **PITCH**.  what does this mean?  what does "YAWPITCH" mean?  what coordinate system is in use here?

lib/src/main/java/org/team100/lib/localization/PoseEstimationHelper.java
```diff
@@ -27,17 +27,20 @@ public class PoseEstimationHelper {
...
-            if (cameraInRobotCoordinates.getRotation().getZ() == Math.PI) {
-                x =-x; 
-            }
-            double y = x * Math.tan(cameraObject.getZ() + cameraInRobotCoordinates.getRotation().getZ());
-            return new Translation2d(x + cameraInRobotCoordinates.getX(), y + cameraInRobotCoordinates.getY());
+            double y = x * Math.tan(cameraObject.getZ());
+            Translation2d cameraRelativeTranslation = new Translation2d(x,y);
+            Rotation2d angleToNote = new Rotation2d(cameraInRobotCoordinates.getRotation().getZ()).plus(cameraRelativeTranslation.getAngle());
+            Translation2d robotRelativeTranslation2dNoOffsets = new Translation2d(angleToNote.getCos()*cameraRelativeTranslation.getNorm(),angleToNote.getSin()*cameraRelativeTranslation.getNorm());
+            return new Translation2d(robotRelativeTranslation2dNoOffsets.getX() + cameraInRobotCoordinates.getX(), robotRelativeTranslation2dNoOffsets.getY() + cameraInRobotCoordinates.getY());
```
truher, line 43: i suspect you can use the "plus" method here.

lib/src/test/java/org/team100/lib/config/SimulatedCameraTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation3d;
+import edu.wpi.first.math.geometry.Transform3d;
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.math.geometry.Translation3d;
+
+public class SimulatedCameraTest {
```
truher, line 16: this test seems like a very simple case: no camera offset and no robot offset.  you should add some non-trivial cases to verify the function of the offsets.

lib/src/test/java/org/team100/lib/localization/TestCameraAngles.java
```diff
@@ -53,4 +53,30 @@ void testCameraAngles4() {
             assertEquals(1, testTranslation1.getY(), 0.0001);
         }
     }
+
+    @Test
+    void testCameraAngles5() {
+        {
```
truher, line 59: why do these extra brackets exist?

lib/src/main/java/org/team100/lib/localization/PoseEstimationHelper.java
```diff
@@ -27,17 +27,20 @@ public class PoseEstimationHelper {
     private static final Telemetry t = Telemetry.get();
     private static final String m_name = Names.name(PoseEstimationHelper.class);
 
-    public static Translation2d cameraRotationToRobotRelative(Transform3d cameraInRobotCoordinates, Rotation3d cameraObject) {
+    public static Translation2d cameraRotationToRobotRelative(Transform3d cameraInRobotCoordinates, Rotation3d yawPitch) {
```
truher, line 30: i don't understand what this method is trying to do at all.  add a long comment explaining it.

lib/src/test/java/org/team100/lib/localization/TestCameraAngles.java
```diff
@@ -53,4 +53,30 @@ void testCameraAngles4() {
...
+            Transform3d e = new Transform3d(0, 0, 1, new Rotation3d(0, Math.toRadians(45), Math.PI));
```
truher, line 60: add comments for each line in these tests explaining what geometry you are verifying

truher (changes requested):

PR #322, opened 2024-03-24
Added DirectPIDControl function to PositionServo, updated climber system

Coolearthsky:


truher (commented):
can you help me understand what this change is for?  it looks like you're adding a method that's the same as the existing one, but uses duty cycle instead of velocity.  why?

comp/swerve100/src/main/java/org/team100/frc2024/LEDSubsystem.java
```diff
@@ -4,9 +4,12 @@
 
 import org.team100.frc2024.motion.shooter.Shooter;
 import org.team100.lib.indicator.LEDIndicator;
+import org.team100.lib.indicator.LEDStrip;
 import org.team100.lib.indicator.LEDIndicator.State;
 import org.team100.lib.localization.VisionDataProvider24;
 
+import com.ctre.phoenix.led.CANdle.LEDStripType;
```
truher, line 11: i think you're using the wrong strip type here, you want something like LEDStrip.strip, i.e. the enum you made below.

comp/swerve100/src/main/java/org/team100/frc2024/LEDSubsystem.java
```diff
@@ -45,29 +48,30 @@ public LEDSubsystem(
 
     @Override
     public void periodic() {
+        m_indicator.setStripSolid(LEDStrip.strip.BACKLIGHT, State.WHITE);
```
truher, line 51: there's no reason to keep resetting the white side to white if it never changes.  is it never going to change?

comp/swerve100/src/main/java/org/team100/frc2024/RobotContainer.java
```diff
@@ -196,10 +197,13 @@ public RobotContainer(TimedRobot robot) throws IOException {
...
+        //TODO get real strips that are front and back
+        List<LEDStrip> strips = new ArrayList<>();
+        for (int i = 0; i< 10;i++) {
+            strips.add(new LEDStrip(16, i*16));
+        }
 
-        LEDStrip strip1 = new LEDStrip(160, 0);
-
-        m_indicator = new LEDIndicator(0, strip1);
+        m_indicator = new LEDIndicator(0, 2, strips.get(0),strips.get(1), strips.get(2), strips.get(3), strips.get(4), strips.get(5), strips.get(6), strips.get(7), strips.get(8), strips.get(9));
```
truher, line 206: i think you really mean to just pass the whole "strips" list as an array, which you can do with:

strips.toArray(new LedStrip[0])

lib/src/main/java/org/team100/lib/indicator/LEDIndicator.java
```diff
@@ -62,21 +63,41 @@ private State(Color color) {
...
+    private final int m_numLEDs;
 
-    public LEDIndicator(int port, LEDStrip... strips) {
+    public LEDIndicator(int port, int numLEDs, LEDStrip... strips) {
+        int numstrips = 0;
+        m_numLEDs = numLEDs;
+        for (LEDStrip strip : strips) {
+            numstrips++;
+        }
+        m_numstrips = numstrips;
```
truher, line 76: i think you mean to say

m_numstrips = strips.length;

lib/src/main/java/org/team100/lib/indicator/LEDIndicator.java
```diff
@@ -62,21 +63,41 @@ private State(Color color) {
...
+        // for (int i = 0; i < m_numstrips; i++) {
+        //     Patterns.solid(leds.get(i * m_numLEDs + 1), blackBuffer, Color.kWhite);
+        // }
         led.setData(buffer);
         led.start();
     }
 
-    public void setStripSolid(int index, State s) {
-        Patterns.solid(leds.get(index), buffer, s.color);
+    public void setStripSolid(LEDStrip.strip state, State s) {
```
truher, line 90: the use of the word "state" in the first parameter is confusing to me.  what 'state' does this describe?

lib/src/main/java/org/team100/lib/indicator/LEDStrip.java
```diff
@@ -5,6 +5,11 @@ public class LEDStrip {
     private final int m_length;
     private final int m_offset;
 
+    public enum strip{
+        BACKLIGHT,
```
truher, line 9: these enum names are very confusing.  don't you just mean "FRONT" and "BACK"?

lib/src/main/java/org/team100/lib/indicator/LEDIndicator.java
```diff
@@ -62,21 +63,41 @@ private State(Color color) {
...
+        if (state.equals(LEDStrip.strip.BACKLIGHT)) {
+            for (int i = 0; i < m_numstrips; i++) {
+                Patterns.solid(leds.get(i * m_numLEDs + 1), buffer, s.color);
+                Patterns.solid(leds.get(i * m_numLEDs + 1), blackBuffer, s.color);
```
truher, line 94: here you are setting a buffer that literally has the word "black" in the name to a color that isn't black, which seems confusing to me.  what are you trying to do?

lib/src/main/java/org/team100/lib/indicator/LEDIndicator.java
```diff
@@ -62,21 +63,41 @@ private State(Color color) {
     private final AddressableLEDBuffer buffer;
     private final AddressableLEDBuffer blackBuffer;
     private final List<LEDStrip> leds = new ArrayList<>();
-    private boolean flashing = false;
+    private boolean flashing;
+    private final int m_numstrips;
+    private final int m_numLEDs;
 
-    public LEDIndicator(int port, LEDStrip... strips) {
+    public LEDIndicator(int port, int numLEDs, LEDStrip... strips) {
```
truher, line 70: the purpose of the "numLEDs" parameter is not clear to me.  what does this mean?

lib/src/main/java/org/team100/lib/indicator/LEDIndicator.java
```diff
@@ -62,21 +63,41 @@ private State(Color color) {
...
+        //     Patterns.solid(leds.get(i * m_numLEDs + 1), blackBuffer, Color.kWhite);
+        // }
         led.setData(buffer);
         led.start();
     }
 
-    public void setStripSolid(int index, State s) {
-        Patterns.solid(leds.get(index), buffer, s.color);
+    public void setStripSolid(LEDStrip.strip state, State s) {
+        if (state.equals(LEDStrip.strip.BACKLIGHT)) {
```
truher, line 91: this logic is very confusing.  what are you trying to do here?

lib/src/main/java/org/team100/lib/indicator/LEDStrip.java
```diff
@@ -5,6 +5,11 @@ public class LEDStrip {
     private final int m_length;
     private final int m_offset;
 
+    public enum strip{
```
truher, line 8: it is very confusing for the enum that represents "which side of the robot is the light facing" to be called "strip".  you should always try to name things according to what they actually mean.

lib/src/main/java/org/team100/lib/indicator/LEDStrip.java
truher, line 1: i think you probably want to make the "which side of the robot the light is on" concept part of the LEDStrip object.

truher (changes requested):
lots of changes required here.

comp/swerve100/src/main/java/org/team100/frc2024/LEDSubsystem.java
truher, line 1: I think this would all be much simpler if you had two indicators at this level, one for front and one for back, instead of a single indicator which need to know about the strip types within it.

lib/src/main/java/org/team100/lib/indicator/LEDIndicator.java
Coolearthsky: I wanted to make it so when we turn on flashing the LEDs on the back don't flash, only the ones on the front

PR #429, opened 2024-07-01
Second der code

Coolearthsky:


lib/src/main/java/org/team100/lib/motion/drivetrain/module/WCPSwerveModule100.java
```diff
@@ -149,10 +148,10 @@ private static PositionServo<Angle100> turningServo(
                 drive);
         PIDController turningPositionController = new PIDController(
                 20, // kP
-                0.06, // kI
+                0, // kI
                 0, // kD
                 dt);
-        turningPositionController.enableContinuousInput(-Math.PI, -Math.PI);
+        turningPositionController.enableContinuousInput(-Math.PI, Math.PI);
```
truher, line 154: woah how did this ever work???

lib/src/main/java/org/team100/lib/motion/components/VelocityServo.java
```diff
@@ -15,6 +15,12 @@ public interface VelocityServo<T extends Measure100> extends Glassy {
      * @param setpoint
      */
     void setVelocity(double setpoint);
+    /**
+     * Velocity in meters/sec or radians/sec depending on T.
+     * 
+     * @param setpoint
```
truher, line 21: if you use @param javadoc you need to list them all.

also you should name the accel setpoint something with the word "accel" in the name, and in the comment you should point out that the units are m/s^2 or rad/s^2

lib/src/main/java/org/team100/lib/motion/components/OutboardVelocityServo.java
```diff
@@ -47,6 +47,13 @@ public void setVelocity(double setpoint) {
         t.log(Level.TRACE, m_name, "Desired setpoint", setpoint);
     }
 
+    @Override
+    public void setVelocity(double setpoint, double setpoint_2) {
+        m_setpoint = setpoint;
+        m_motor.setVelocity(setpoint, setpoint_2, 0);
+        t.log(Level.TRACE, m_name, "Desired setpoint", setpoint);
```
truher, line 54: is there a reason you're not loggin gthe accel setpoint?

comp/swerve100/src/main/java/org/team100/frc2024/motion/shooter/DrumShooter.java
```diff
@@ -224,7 +224,7 @@ public boolean atVelocitySetpoint() {
         }
         double leftError = leftVelocity.getAsDouble() - kLeftRollerVelocity;
         double rightError = rightVelocity.getAsDouble() - kRightRollerVelocity;
-        return (Math.abs(leftError) < 10) && (Math.abs(rightError) < 10);
+        return (Math.abs(leftError) < .5) && (Math.abs(rightError) < 0.5);
```
truher, line 227: always use a leading zero

truher (changes requested):
i left a few comments.  the most important ones are (a) make sure the "experiment disabled" case is identical to the code as it was before, and (b) don't represent angular velocity as a rotation.

comp/vision/tag_finder24.py
```diff
@@ -191,32 +191,32 @@ def __init__(self, serial, camList):
...
-            (point2D, _) = cv2.projectPoints(
-                object_points, rvec, tvec, camera.mtx, camera.dist
-            )
-            if (
-                point2D[0][0][0] > 0
-                and point2D[0][0][0] < 1456
-                and point2D[0][0][1] > 0
-                and point2D[0][0][1] < 1088
-            ):
+        # potentialTags = self.estimatedTagPose.get()
```
truher, line 194: big swaths of commented-out code are bad.  maybe you could make a copy of tag_finder24 with this stuff in it?

truher (approved):
i'm going to merge this as is but you might still want to make another file to work with in python

PR #452, opened 2024-07-20
demo bot code

Coolearthsky:


lib/src/main/java/org/team100/lib/motion/components/OutboardLinearPositionServo.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.OptionalDouble;
+
+import org.team100.lib.controller.State100;
+import org.team100.lib.motion.LinearMechanism;
+import org.team100.lib.profile.Profile100;
+import org.team100.lib.telemetry.SupplierLogger;
+import org.team100.lib.telemetry.Telemetry.Level;
+
+/**
+ * Position control using duty cycle feature of linear mechanism
```
truher, line 12: this is another example of cutting and pasting stuff without reading it

lib/src/main/java/org/team100/lib/motion/components/OutboardLinearPositionServo.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+public class OutboardLinearPositionServo implements LinearPositionServo {
```
truher, line 14: the word "outboard" in our code means "using the feedback controller in the motor controller hardware" which is not what you're doing here.

comp/swerve100/src/main/java/org/team100/frc2024/motion/climber/ClimberPosition.java
```diff
@@ -49,8 +49,8 @@ public void initialize() {
 
     @Override
     public void execute() {
-        m_leftServo.setPosition(m_goalM);
-        m_rightServo.setPosition(m_goalM);
+        m_leftServo.setPosition(m_goalM, 0);
```
truher, line 52: this sort of change is why you shouldn't do it this way.

lib/src/main/java/org/team100/lib/util/CollectionUtil.java
```diff
@@ -0,0 +1,112 @@
...
+        SimulatedBareMotor driveMotor = new SimulatedBareMotor(parent, 5);
+        // simulated gearing is 2 meter wheel, 1:1, so rad/s and m/s are the same.
+        return new SimpleLinearMechanism(
+                driveMotor,
+                new SimulatedBareEncoder(parent, driveMotor),
+                1,
+                2);
+    }
+
+    private static LinearMechanism getMechanismNoEncoder(SupplierLogger parent, Neo550CANSparkMotor motor, double gearRatio, double wheelDiameterM) {
```
truher, line 105: what does "no encoder" mean here?

truher (changes requested):
lots of comments

lib/src/main/java/org/team100/lib/commands/drivetrain/manual/ManualWithFullStateHeading.java
```diff
@@ -132,7 +132,15 @@ public FieldRelativeVelocity apply(SwerveState state, DriverControl.Velocity twi
         if (m_thetaSetpoint == null) {
             // TODO: to avoid overshoot, maybe pick a setpoint that is feasible without
             // overshoot?
-            updateSetpoint(yawMeasurement, yawRate);
+            double accel;
```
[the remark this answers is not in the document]
Coolearthsky: I was hoping this could fix the todo by creating a setpoint which changes slowly with the max acceleration of the robot, which is what hte code I added should do.

lib/src/main/java/org/team100/lib/motion/components/GravityServo.java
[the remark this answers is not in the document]
Coolearthsky: This is for the pivot shooter.

studies/python_deployment/app/tag_detector.py
```diff
@@ -153,7 +153,8 @@ def analyze2(self, metadata: dict[str, Any], buffer: mmap) -> None:
         delay_ns: int = Timer.time_ns() - sensor_midpoint_ns
         delay_us = delay_ns // 1000
 
-        self.network.vision_nt_struct.set(blips, ntcore._now() - delay_us)
+        self.network.vision_nt_struct.set(blips)
```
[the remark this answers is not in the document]
Coolearthsky: The set method takes 1 argument, I am not sure how this worked at all, as it just crashes every time I run this, in the next line I added self.network.vision_capture_time_ms.set(ntcore._now() - delay_us) to still put up the capture time.

lib/src/main/java/org/team100/lib/motion/components/LinearPositionServo.java
```diff
@@ -21,7 +21,29 @@ public interface LinearPositionServo extends Glassy {
      * 
      * @param goalM
      */
-    void setPosition(double goalM);
+     /**
```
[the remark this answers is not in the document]
Coolearthsky: Fixed

lib/src/main/java/org/team100/lib/motion/components/OutboardLinearPositionServo.java
Coolearthsky: Fixed

studies/python_deployment/app/tag_detector.py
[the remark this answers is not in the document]
truher: I think you're not aware of what the time field does: it synchronizes the server time to the frame time, with correction for network latency. if it doesn't work, don't just turn it off, figure out why.  I can help you work through this if you want.

lib/src/main/java/org/team100/lib/commands/drivetrain/manual/ManualWithFullStateHeading.java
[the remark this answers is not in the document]
truher: i think that comment should actually be about adjusting the *goal*, not the setpoint.  the setpoint needs to be whatever the current robot state is, if you want to avoid some sort of jerky transition.

lib/src/main/java/org/team100/lib/motion/components/LinearPositionServo.java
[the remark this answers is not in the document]
truher: it looks like you're adding gravity compensation here but then you never use it.  gravity compensation might be useful for, say, a vertically oriented elevator.  but i don't see that here.

lib/src/main/java/org/team100/lib/config/Feedforward100.java
```diff
@@ -35,7 +35,7 @@ public class Feedforward100 {
     private final double kDS;
     private final double staticFrictionSpeedLimit;
 
-    Feedforward100(double kV, double kA, double kSS, double kDS, double staticFrictionSpeedLimit) {
+    private Feedforward100(double kV, double kA, double kSS, double kDS, double staticFrictionSpeedLimit) {
```
truher, line 38: i think you're not aware of what the "default" visibility does.  if a method is neither "public" nor "private" that means that it is "package private" and can be seen by other classes in the same package.  in this case, that's "org.team100.lib.config," and the most interesting other class in that package is the unit test of this class.  so you ought to put this back like it was.

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
```diff
@@ -25,7 +25,9 @@
  * Sanjans version
  */
 public class DriveWithProfile2 extends Command100 {
-    private static final double kTolerance = 0.1;
+    private static final double kRotationToleranceRad = Math.PI/16;
```
truher, line 28: this is ten degrees, which feels like a lot?

lib/src/main/java/org/team100/lib/motor/Neo550CANSparkMotor.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public double kROhms() {
+        return 0.12;
+    }
+
+    @Override
+    public double kTNm_amp() {
+        return 0.3944;
```
truher, line 34: where did this number come from?  0.39 Nm per amp is an enormous number, it means at the stall current of 100 amps the motor will be making 39 Nm which is like the tightening torque on a 5/16" bolt, i.e. huge compared to the output of a neo 550.

lib/src/main/java/org/team100/lib/config/Feedforward100.java
```diff
@@ -47,7 +47,11 @@ public static Feedforward100 makeNeo() {
         return new Feedforward100(0.122, 0.000, 0.100, 0.065, 0.5);
     }
 
-    public static Feedforward100 makeArmPivot() {
+    public static Feedforward100 makeNeo550() {
+        return new Feedforward100(0.32, 0, 0, .07, 0);
```
truher, line 51: where did the number 0.32 come from?  0.32 volt-sec per revolution means it's only going about 3400 rpm at 12 volts.

studies/demoBot/src/main/java/org/team100/frc2024/shooter/ShooterWithVerticalPivot.java
```diff
@@ -0,0 +1,72 @@
...
+
+import edu.wpi.first.wpilibj.Servo;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class ShooterWithVerticalPivot extends SubsystemBase implements Glassy{
+    
+    private final DrumShooter m_shooter;
+    private final SupplierLogger m_logger;
+    private final OutboardGravityServo m_pivot;
+    private final Servo m_indexer;
```
truher, line 18: now you're not using the indexer subsystem at all, you're using a WPILib Servo?  i'm confused about how this is supposed to work.  you're not supposed to use both the setAngle() and setSpeed() methods of the same servo; they apply to different kinds of hardware.

studies/demoBot/src/main/java/org/team100/frc2024/shooter/indexer/IndexerSubsystem.java
truher, line 1: i think you should do the indexer subsystem differently.  it looks like you want to control it in one of two modes: either positional mode or velocity mode.  you should make those modes different commands, and put the positional controller in the command.  see the 2024 Climber subsystem and related commands for examples to follow.

truher (changes requested):

studies/demoBot/src/main/java/org/team100/frc2024/drivetrain/TankDriveSubsystem.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public String getGlassName() {
+        return "Tank Drive Subsystem";
+    }
+
+    @Override
+    public void periodic() {
+      m_logger.logOptionalDouble(Level.TRACE, "Left Drive Velocity M_S",() -> m_modules.getSpeeds()[0]);
```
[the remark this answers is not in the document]
Coolearthsky: It comes from an encoder which returns optional double

studies/demoBot/src/main/java/org/team100/frc2024/shooter/ShooterCollection.java
```diff
@@ -0,0 +1,69 @@
...
+import org.team100.lib.config.Identity;
+import org.team100.lib.motion.components.GravityServo;
+import org.team100.lib.motion.components.LinearVelocityServo;
+import org.team100.lib.motion.components.OutboardAngularPositionServo;
+import org.team100.lib.motor.MotorPhase;
+import org.team100.lib.telemetry.SupplierLogger;
+import org.team100.lib.util.CollectionUtil;
+import edu.wpi.first.wpilibj.Servo;
+
+public class ShooterCollection {
```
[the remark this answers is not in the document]
Coolearthsky: I thought the main point of the class was to create a simulated example aside from just the shooter

studies/python_deployment/app/tag_detector.py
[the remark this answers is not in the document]
Coolearthsky: I am confused I might need some help in person on this

lib/src/main/java/org/team100/lib/motor/Neo550CANSparkMotor.java
Coolearthsky: I might need some help on this, how I did it was I took the torque NM from the graph and devided it by the amps.

lib/src/main/java/org/team100/lib/config/Feedforward100.java
Coolearthsky: This was just what I got from testing it for a few minutes, this number is likely not acurate, i'll add a todo.

studies/demoBot/src/main/java/org/team100/frc2024/shooter/ShooterWithVerticalPivot.java
Coolearthsky: I probably will need to test this once I get the actual build, I am not sure how it would work. The indexer subsystem is not being used but I created it in case we need to pivot (no pun intended) from a servo into a neo550 or other motor.

studies/demoBot/src/main/java/org/team100/frc2024/shooter/indexer/IndexerSubsystem.java
Coolearthsky: Fixed

PR #508, opened 2024-10-08
Demo bot update

Coolearthsky:


lib/src/main/java/org/team100/lib/hid/DriverControl.java
```diff
@@ -56,6 +56,10 @@ default Velocity velocity() {
         return new Velocity(0, 0, 0);
     }
 
+    default double yAxis() {
```
truher, line 59: the pattern here is to name things according to what they accomplish, so this would be something like "shooterPivot"

truher (changes requested):

PR #509, opened 2024-10-11
Work on driveWithProfile2

Coolearthsky:


lib/src/main/java/org/team100/lib/profile/TrapezoidProfile100.java
```diff
@@ -67,7 +67,8 @@ public TrapezoidProfile100(double maxVel, double maxAccel, double tolerance) {
     }
 
     public TrapezoidProfile100 scale(double s) {
-        return new TrapezoidProfile100(m_maxVelocity, s * m_maxAcceleration, m_tolerance);
+        m_maxAcceleration *= s;
+        return new TrapezoidProfile100(m_maxVelocity, m_maxAcceleration, m_tolerance);
```
truher, line 71: I think I see the issue, you're trying to make the profile immutable?  is this better though?  maybe  make it all immutable somehow?

truher (changes requested):

PR #524, opened 2024-10-19
Add instructions on setting up the RoboRIO

dobromirmontauk:
Initial PR that adds more instructions on how to get set up with the `all24` codebase, including getting it running on the Simulator and the RoboRIO.


comp/swerve100/README.md
```diff
@@ -2,4 +2,70 @@
...
+TODO.
+
+The rest of the instructions assume that you have the following setup:
+1. Git `Team100/all24` repo checked out.
+1. Custom WPI VSCode with extensions installed.
+1. `swerve100` workspace opened.
+
+## Building the RoboRIO Code
+Use the WPI extensions to run the `gradlew` build:
+1. Press `CTRL + SHIFT + P` to get the VSCode Command Mode.
```
truher, line 19: i usually click on the little "W" icon in the upper left corner, which brings up the WPI commands (like build and deploy)

comp/swerve100/README.md
```diff
@@ -2,4 +2,70 @@
...
+## Simulating the RoboRIO
+Use the WPI extensions to kick off the Simulator. Good overall instructions are available [here](https://docs.wpilib.org/en/stable/docs/software/wpilib-tools/robot-simulation/simulation-gui.html). 
+1. Press `CTRL + SHIFT + P` to get the VSCode Command Mode.
+1. Search for `Simulate Robot Code`
+1. Select `swerve100` when it asks.
+1. It will ask you what simulations to run. Make sure `Sim GUI` is selected, and nothing else.
+1. You should get the robot simulation GUI, which looks like this:
+![Simulation GUI](readme_img/sim_gui.png)
+1. Make sure you can debug your code in simulation.
+    1. If you aren't familiar with debugging in VSCode, read this: TODO.
```
truher, line 39: we actually don't use the debugger very much, i think mostly because you can't run it on the real robot (as far as i know), and even if you could, all the behavior is so timing-dependent that it wouldn't be useful.  instead we rely on logging and unit tests.

truher (approved):
looks good.  i left a couple of comments which you're free to ignore.  :-)

truher:
did you mean to leave this PR in "draft"?

comp/swerve100/README.md
dobromirmontauk: OK interesting. I guess I'm not used to coding w/o access to the debugger :)

PR #528, opened 2024-10-28
Visual servoing

Coolearthsky:


raspberry_pi/app/camera/real_camera.py
```diff
@@ -230,7 +230,7 @@ def __size_from_model(model: Model) -> Size:
 
             case Model.GS:
                 return Size(
-                    sensor_width=1456, sensor_height=1088, width=1456, height=1088
```
truher, line 233: why are you using 1408 instead of 1456 here?

truher (approved):

PR #545, opened 2024-11-06
Visual servoing

Coolearthsky:


lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
```diff
@@ -82,14 +83,14 @@ public void initialize() {
         ySetpoint = m_swerve.getState().y();
         thetaSetpoint = m_swerve.getState().theta();
 
-        Optional<Pose2d> opt = m_fieldRelativeGoal.get();
+        Optional<Pose2d> opt = m_fieldRelativeGoalSupplier.get();
```
truher, line 86: it looks to me like this command only does the "coordinated profile" thing if it happens to run with a goal the first time it is "initialized".  also, note that the coordination scaling changes the profiles, so if you press the button over and over, the profiles will get slower and slower.  maybe take out all this initialization coordination stuff until the basic uncoordinated profile is working perfectly?

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
```diff
@@ -112,22 +113,23 @@ public void execute() {
...
+            if (m_fieldRelativeGoal == null) 
             return;
+        } else {
+            m_fieldRelativeGoal = opt.get();
         }
-        Pose2d fieldRelativeGoal = opt.get();
-
         Rotation2d bearing = new Rotation2d(
-                Math100.getMinDistance(measurement, fieldRelativeGoal.getRotation().getRadians()));
+                Math100.getMinDistance(measurement, m_fieldRelativeGoal.getRotation().getRadians()));
```
truher, line 124: note the word "bearing" means "angle from point A to point B" but the expression here appears to return simply the angular component of the goal pose itself, which isn't the same thing.  what do you mean this to mean?

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
```diff
@@ -159,7 +161,7 @@ public void execute() {
 
     @Override
     public boolean isFinished() {
-        if (!m_fieldRelativeGoal.get().isPresent()) return true;
+        if (!m_fieldRelativeGoalSupplier.get().isPresent() && m_fieldRelativeGoal == null) return true;
```
truher, line 164: don't you want the command to keep going even when it loses sight of the target?

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
```diff
@@ -112,22 +113,23 @@ public void execute() {
         Rotation2d currentRotation = m_swerve.getState().pose().getRotation();
         // take the short path
         double measurement = currentRotation.getRadians();
-        Optional<Pose2d> opt = m_fieldRelativeGoal.get();
+        Optional<Pose2d> opt = m_fieldRelativeGoalSupplier.get();
         if (opt.isEmpty()) {
+            if (m_fieldRelativeGoal == null)
```
truher, line 118: this logic seems to make the command end one cycle after the target is lost.

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
```diff
@@ -55,7 +56,7 @@ public DriveWithProfile2(
             SwerveDriveSubsystem drivetrain,
             HolonomicFieldRelativeController controller,
             SwerveKinodynamics limits) {
-        m_fieldRelativeGoal = fieldRelativeGoal;
+        m_fieldRelativeGoalSupplier = fieldRelativeGoal;
         m_swerve = drivetrain;
         m_controller = controller;
         m_limits = limits;
```
truher, line 62: these profile limits are unattainably fast, aren't they?  did you try making the profiles a lot slower, as i suggested?  "first make it perfect, then make it fast"?

truher (changes requested):
hey, i'd like to get this merged, and generally i'd like this project to hurry up and get finished, but i have some comments i'd like you to see.  is this the final code from today?  it seems like you were doing things that i don't see here.

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
Coolearthsky: This part is mainly to prevent the case where when I press the button no note is identified having some random result. This will only happen if no object has ever been detected since the robot has been booted due to the fact it only is null upon initialization of the class and is changed upon the first initialization of the command given a pose is given.

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
Coolearthsky: See other comment about null.

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
Coolearthsky: I figured the whole calculation might take too long to run each cycle, however I could be wrong about that, so i'll change that, the profile does not get slower and slower due to the fact the scaling is undone in the end function.

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
Coolearthsky: Fixed

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
Coolearthsky: Fixed

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
truher: tbh I would suggest leaving the coordination part out until the rest works perfectly.

truher: if you want to keep it, definitely don't do the "undo at the end" thing.  just don't overwrite the original.

lib/src/main/java/org/team100/lib/commands/drivetrain/DriveWithProfile2.java
truher: hm, I think it's not doing what you think it's doing.  maybe start with something simple that can work perfectly.

